Issue #18, opened 2020-06-24
Implement mock UPNP and MDNS scanning


cdaly333:
Add UPNP and MDNS scan configs as well as functionality to scan mock devices using these configurations

cdaly333:
Closing because scanning is no longer part of the tested flow

PR #28, opened 2020-06-26
Add static MockLocalHomePlatform instance for reference from App instances

### LINKED ISSUE #19: Make stubs interact with an instance of a Mock Local Home Platform
When the Bundled JavaScript calls the stub constructors, we need to inject a reference to a running Local Home Platform so the stub instances can properly communicate with it.

cdaly333:
**#29 depends on this PR**

This resolves #19 based on the singleton pattern.  It adds a static instance of `MockLocalHomePlatform` so an arbitrary `App` instance can link with it.

**Example:**
1. In a test, developer would get the singleton instance of `MockLocalHomePlatform` and call `initializeRadio()` to pass in scan configuration and inject the `MockNetwork`.  `initializeRadio()` creates a `DeviceManager` based on the `MockNetwork`.
2. `App` constructor is called.  The `App` instance injects itself into the singleton `MockLocalHomePlatform`.  This is so the `MockLocalHomePlatform` can access the handlers (implemented in #29).
3. Now the `MockLocalHomePlatform` can call access the proper handlers when needed.  When `App.getDeviceManager()` is called, the `App` returns the `DeviceManager` attached to the `MockLocalHomePlatform` singleton instance.

**Notes:**
- Alternative: `App` could be a singleton, and `MockLocalHomePlatform` could create and inject a `DeviceManager`.  I don't see any significant advantage/disadvantage of this alternative
- Not sure if singleton is the appropriate pattern for this


cdaly333:
Removed WIP tag

platform/mock-local-home-platform.ts
```diff
@@ -38,28 +39,58 @@ export class UDPScanConfig implements ScanConfig {
...
+  private deviceManager: smarthome.DeviceManager;
   private localDeviceIds: string[] = [];
 
-  constructor(network: MockNetwork, udpScanConfigs: UDPScanConfig[]) {
-    this.network = network;
+  private constructor() {
+    this.setupUDP();
+  }
+
+  public initializeRadio(
```
Fleker, line 57: Will this need to change to accommodate other transport protocols aside from UDP?

platform/device-manager.ts
```diff
@@ -3,24 +3,18 @@
...
   deviceId: string;
   error: boolean;
   commands: smarthome.DataFlow.TcpRequestData[] = [];
+  private mockNetwork: MockNetwork;
+
+  constructor(mockNetwork: MockNetwork) {
+    this.mockNetwork = mockNetwork;
+  }
+
   public send(
```
Fleker, line 18: In Typescript, do methods need to be marked explicitly as public?

platform/mock-local-home-platform.ts
cdaly333: Yeah, I will need to update this function once I address #18.  I plan to use optional parameters.

platform/mock-local-home-platform.ts
```diff
@@ -38,28 +39,58 @@ export class UDPScanConfig implements ScanConfig {
...
-  private network: MockNetwork;
+  private mockNetwork: MockNetwork;
+  private app: smarthome.App;
+  private deviceManager: smarthome.DeviceManager;
   private localDeviceIds: string[] = [];
 
-  constructor(network: MockNetwork, udpScanConfigs: UDPScanConfig[]) {
-    this.network = network;
+  private constructor() {
+    this.setupUDP();
```
cdaly333: Removed this line, since `initializeRadio` replaces this call.

platform/device-manager.ts
cdaly333: The default modifier is `public`, so I don't really need to, no.  I've been adding it out of habit since I'm not used to TypeScript, but I could remove them (I think there's an eslint rule I can add).  Might slightly help others not used to TypeScript to leave it in.

platform/device-manager.ts
```diff
@@ -3,24 +3,18 @@
  * TODO(cjdaly) integrate with Mock Local Home Platform and Mock Network
```
proppy, line 3: should we review the common trunk of changes as part of #29 or #28?

It may help to rebase the change on top of each others branches (rather than having both against master) if they are truely dependent on each other.

cdaly333: Yeah they are dependent on each other I will look into this.

cdaly333: I'm assuming I won't be able to open a PR on this repo with one branch targeting another branch if both of the branches exist on a fork (haven't tested this).  I added a reference to #29 to the description of this PR.  For now I think we should review the common changes on this PR.  I'll continue looking into the best way to do this though.

cdaly333: For what it's worth, I can confirm I can't create a PR on this repo with both branches on my fork.

cdaly333:
Closing as #29 was merged which included these changes

Issue #21, opened 2020-06-25
Export functionality as node module that can be imported from other projects


cdaly333:
Per [proppy's comment, ](https://github.com/googleinterns/smart-home-local/pull/13#discussion_r443987701), the stubs should be packaged as a library that can be imported into the [Local Home SDK sample](https://github.com/actions-on-google/smart-home-local), rather than using this node project as the entry point.  This is related to Issue #7.

Issue #27, opened 2020-06-26
Simplify Identify functionality by removing scan/broadcast


cdaly333:
Instead of making the developer implement device radio responses, consider if that degree of freedom is necessary.  It may be better to short circuit that flow and simply manually trigger messages from a test.

Issue #23, opened 2020-06-26
Fully implement DeviceManager stub


cdaly333:
-Add assert to check which commands were sent
-Route DeviceManager.send() calls to corresponding mock network messaging components.

PR #29, opened 2020-06-27
Implement Identify flow and App stub

### LINKED ISSUE #15: Implement IDENTIFY flow functionality
Implement the Local Home Platform broadcast/scan functionality.  Have a mock device respond to the virtual broadcast with an IDENTIFY request, and have the mock Local Home Platform process this request properly using the bundled JavaScript app

### LINKED ISSUE #24: Fully implement App stub
App stub currently implements the `smarthome.App` interface but functions just return.
Need to fill in these implementations with proper behavior

cdaly333:
**Depends on #28**

This PR implements the `IDENTIFY` flow described in the design document.
Briefly, that is:

1. Scan configuration is set
2. Scan is triggered
3. A UDP broadcast packet is sent out by `MockLocalHomePlatform`
4. A mock device receives the broadcast packet, and responds with a `deviceId`
5. The `MockLocalHomePlatform` sends an `IdentifyRequest` to the fulfillment app with encoded scan data
6. The `MockLocalHomePlatform` receives the `IdentifyResponse` from fulfillment app
7. The `MockLocalHomePlatform` saves the `localDeviceId` in memory.  Mock device is now conceptually connected with the MockLocalHomePlatform.

**Changes:**
- The `AppStub` was fully implemented.  `AppStub` now saves the passed in handlers.
- `identifyHandler` has a union return type of `IdentifyResponse | Promise<IdentifyResponse>` , so methods were added to `MockLocalHomePlatform` to handle both cases asynchronously.
- The sample test was updated to wait for the correct `deviceId` to be registered.  This was achieved by adding an `onNewDeviceIdRegistered` event to `MockLocalHomePlatform`.

**Issues**:
- resolves #24
- resolves #15

**Notes**:
- UDP is curently the only radio / scan configuration that is supported (see #18)
- Fulfillment app is currently integrated as a bundled app.  This will chance once this project is packaged as node package. (see #21)
- Test will timeout if no `localDeviceId` ever registers.
- Tagging @limlam


platform/device-manager.ts
```diff
@@ -3,24 +3,18 @@
...
-  }
-
+export class DeviceManagerStub implements smarthome.DeviceManager {
   deviceId: string;
   error: boolean;
   commands: smarthome.DataFlow.TcpRequestData[] = [];
+  private mockNetwork: MockNetwork;
+
+  constructor(mockNetwork: MockNetwork) {
+    this.mockNetwork = mockNetwork;
```
proppy, line 15: how is this being used?

platform/mock-local-home-platform.ts
```diff
@@ -69,38 +108,56 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
             structureData: {},
             params: {},
           },
         },
       ],
       devices: [],
-      rinfo,
+    };
+
+    this.processIntentResponse<smarthome.IntentFlow.IdentifyResponse>(
```
proppy, line 153: wouldn't this work?
```
Promise.resolve(this.app.identifyHandler(identifyRequest)).then(...)
```
If `identifyHandler` result is a Promise, it will cascade and if it's a result it will wrap it in a promise.

platform/stub-setup.ts
```diff
@@ -1,13 +1,15 @@
...
  */
-import { DeviceManagerStub } from './device-manager';
 import { AppStub } from './smart-home-app';
 
 (global as any).smarthome = {
   App: AppStub,
-  DeviceManager: DeviceManagerStub,
+  Intents: {
+    IDENTIFY: 'action.devices.IDENTIFY',
+    EXECUTE: 'action.devices.EXECUTE',
```
proppy, line 11: REACHABLE?

platform/stub-setup.ts
```diff
@@ -1,13 +1,15 @@
...
+ * Injects the App stub into the global scope.
+ * Bundled HomeApp cannot be loaded until this global is set.
  */
-import { DeviceManagerStub } from './device-manager';
 import { AppStub } from './smart-home-app';
 
 (global as any).smarthome = {
   App: AppStub,
-  DeviceManager: DeviceManagerStub,
+  Intents: {
```
proppy, line 9: can we type this with https://github.com/actions-on-google/local-home-sdk/blob/9840f62601fdefcf9144de4f0115e9a71c70e07f/execution-only.d.ts#L45 ?

package.json
```diff
@@ -13,12 +13,13 @@
...
     "@typescript-eslint/eslint-plugin": "^3.3.0",
     "@typescript-eslint/parser": "^3.3.0",
-    "ts-node": "^8.10.2",
     "ava": "^2.1.0",
+    "cbor": "^5.0.2",
     "eslint": "^7.2.0",
     "eslint-config-google": "^0.14.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-prettier": "^3.1.4",
+    "ts-node": "^8.10.2",
```
proppy, line 22: newer version of ava shouldn't need this, see:
https://github.com/avajs/typescript

test/test.ts
```diff
@@ -3,19 +3,21 @@
  */
 /// <reference types="@google/local-home-sdk" />
 import test from 'ava';
+import cbor from 'cbor';
```
proppy, line 6: maybe we could use something simpler (and built in) for the test?

test/test.ts
```diff
@@ -75,8 +78,20 @@ test('udp-device-connects', (t) => {
     scanConfig.broadcastAddress
   );
 
+  loadHomeApp('../home-app/bundle');
```
proppy, line 81: what about including the test app implemention inline rather than loading it from the filesystem?

test/test.ts
```diff
@@ -75,8 +78,20 @@ test('udp-device-connects', (t) => {
...
+
+  const connectedDeviceId = new Promise<string>((resolve, reject) => {
+    mockLocalHomePlatform.addOnNewDeviceIdRegistered(
```
proppy, line 84: what about making this an async public getter of the platform rather than a push-based listener?

platform/mock-local-home-platform.ts
```diff
@@ -38,28 +40,65 @@ export class UDPScanConfig implements ScanConfig {
...
 export class MockLocalHomePlatform implements MockUDPListener {
+  //  Singleton instance
+  private static instance: MockLocalHomePlatform;
+
   private udpScanConfigs: UDPScanConfig[] = [];
-  private network: MockNetwork;
+  private mockNetwork: MockNetwork;
+  private app: AppStub;
+  private deviceManager: smarthome.DeviceManager;
   private localDeviceIds: string[] = [];
```
proppy, line 52: this should probably be a map (the platform should store the association between deviceId and localDeviceId)

test/test.ts
```diff
@@ -25,12 +27,13 @@ test('udp-device-connects', (t) => {
...
+  // Mock a UDP Network
   const mockNetwork = new MockNetwork();
 
   // Mock the Local Home Platform
-  const mockLocalHomePlatform = new MockLocalHomePlatform(mockNetwork, [
-    scanConfig,
-  ]);
+  const mockLocalHomePlatform = MockLocalHomePlatform.getInstance();
+
+  mockLocalHomePlatform.initializeRadio(mockNetwork, [scanConfig]);
```
proppy, line 36: what about simplifying this (#27) rather than introducing a new public interface element.

platform/device-manager.ts
cdaly333: This would be part of #23 to allow the fulfillment app to utilize the `MockNetwork`.  I'll remove from this PR.

platform/mock-local-home-platform.ts
cdaly333: That makes a lot more sense.  Will make this change

platform/stub-setup.ts
cdaly333: Will add.

platform/stub-setup.ts
cdaly333: Yeah, I was wondering about this.  Since I'm setting the `smarthome` namespace I can't access `smarthome.Intents` from the SDK typings.  I guess I could declare it `Intents: smarthome.Intents`

package.json
cdaly333: Will remove

test/test.ts
cdaly333: I added this back because it works with the [sample](https://github.com/actions-on-google/smart-home-local).  I guess in the long run I should keep all the tests in this file more generic, but I think that falls into issue #1

test/test.ts
cdaly333: I will make this change in #30

test/test.ts
cdaly333: That's a good idea.  Will add

platform/mock-local-home-platform.ts
cdaly333: will add

test/test.ts
cdaly333: Will address in #27

package.json
cdaly333: Will address in #30

platform/mock-local-home-platform.ts
```diff
@@ -69,38 +108,56 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
       scanConfig.broadcastAddress,
       scanConfig.listenPort,
       scanConfig.broadcastAddress
     );
   }
+  private isPromise<RS>(response: RS | Promise<RS>): response is Promise<RS> {
+    return (response as Promise<RS>).then !== undefined;
+  }
+
+  private processIntentResponse<RS>(response: RS | Promise<RS>): Promise<RS> {
```
Fleker, line 123: If you mark this function as an `async`, you can `await` the `response` with [minimal side effects](https://stackoverflow.com/questions/55262996/does-awaiting-a-non-promise-have-any-detectable-effect) regardless of whether it is a Promise or not. It'll clean up this code.

platform/mock-local-home-platform.ts
Fleker: If you mark the function as `async` you can `await` and make the code look more synchronous.

platform/smart-home-app.ts
```diff
@@ -3,30 +3,46 @@
...
   }
+
   getDeviceManager(): smarthome.DeviceManager {
-    return;
+    return MockLocalHomePlatform.getInstance().getDeviceManager();
   }
+
   listen(): Promise<void> {
-    return;
+    return Promise.resolve();
```
Fleker, line 27: Is this method to be implemented later?

platform/smart-home-app.ts
```diff
@@ -3,30 +3,46 @@
...
+
+  public onExecute(executeHandler: smarthome.IntentFlow.ExecuteHandler): this {
+    this.executeHandler = executeHandler;
+    return this;
   }
-  onIdentify(handler: smarthome.IntentFlow.IdentifyHandler): this {
-    return;
+
+  public onIdentify(
+    identifyHandler: smarthome.IntentFlow.IdentifyHandler
```
Fleker, line 36: Is your linter checking the number of columns per-line? Does this go over the limit?

platform/stub-setup.ts
cdaly333: Guess I can't do this since it's a type.  Seems redundant that I need to redefine Intents here.

platform/smart-home-app.ts
cdaly333: I think it's currently at the default line limit for `eslint`.  I'll fix this with the dependency fixes in #30

platform/device-manager.ts
cdaly333: Done

platform/mock-local-home-platform.ts
cdaly333: Neat. Done.

platform/stub-setup.ts
cdaly333: Done. Referenced sdk typing.

test/test.ts
cdaly333: Done.

platform/mock-local-home-platform.ts
cdaly333: Done

platform/mock-local-home-platform.ts
cdaly333: used [Proppy's solution](https://github.com/googleinterns/smart-home-local/pull/29#discussion_r447494848) with `async` and `await`

platform/stub-setup.ts
proppy: Something like this:
```
const smarthomeStub: {
  App: smarthome.App
  DeviceManager: smarthome.DeviceManager
  Intents: {[key in keyof typeof smarthome.Intents]: string}
} = {
  App: AppStub,
  DeviceManager: DeviceManagerStub,
  Intents: {
    IDENTIFY: 'action.devices.IDENTIFY',
    EXECUTE: 'action.devices.EXECUTE',
  }
};
```
Could introduce a little bit more typing and help you catch if you're forgetting any enum keys. (This wouldn't let you typecheck against enums values since those seems to be opaque by design https://github.com/microsoft/TypeScript/issues/17690#issuecomment-321319291)

This might be worth raising an issue so that the SDK typing expose this using a proper union type instead.

cdaly333: Thanks, I added the typing.  Can you elaborate on the use of a "proper union type"? Are you suggesting the SDK expose `smarthome.Intents` as a union type of string literals?

platform/smart-home-app.ts
cdaly333: Good call.  I suppose the intended behavior is that `listen()` notifies the Local Home Platform that the Local Fulfillment app has set all the handlers and is ready to receive requests.  I added a flag so that developers can assert `listen()` was called.  Also added an example of this flag being asserted.

platform/mock-local-home-platform.ts
```diff
@@ -81,6 +82,14 @@ export class MockLocalHomePlatform implements MockUDPListener {
     });
   }
 
+  public getIsHomeAppReady(): boolean {
```
Fleker, line 85: Does this need to be a `get`? As it's a boolean, having `isHomeAppReady` would imply a getter for a boolean.

cdaly333: I changed the boolean name and the getter to `homeAppReady` and `isHomeAppReady`, respectively.  Let me know if this isn't what you meant.

cdaly333:
If the common changes from #28 also have been approved we can just merge this PR and close #28.  Will need to find a workaround for these dependent PRs that makes it more manageable for reviewers.

Issue #26, opened 2020-06-26
Use built in EventEmitter instead of custom Listener


cdaly333:
Instead of using custom listener pattern, implement [existing node listener](https://nodejs.org/api/events.html)

cdaly333:
Closing since current implementation doesn't rely heavily on events.

PR #30, opened 2020-06-29
Export stubs as node module for usage from external Local Fulfillment node projects

### LINKED ISSUE #21: Export functionality as node module that can be imported from other projects

### LINKED ISSUE #22: Improve isolation of bundled JavaScript
Implement node.js vm module to isolate bundled javascript.  See [proppy's comments](https://github.com/googleinterns/smart-home-local/pull/13#discussion_r445367302)

cdaly333:
Refactored the project to be used as an external node module rather than needing to import the Local Fulfillment App into the project

**Main Changes:**
- Implemented [Google's TypeScript Style Guide](https://github.com/google/gts).  This included moving everything into a `src` folder, among other changes.
- Switched from `eslint` to `tslint` for consistency across [Local Home SDK](https://github.com/actions-on-google/local-home-sdk) and [SDK Sample](https://github.com/actions-on-google/smart-home-local) repos.
- Added a few null checks and placeholder return statements to pass `tslint` rules
- Remove `bundle.js` from project
- Remove `ava` test script from `package.json`.  `test.ts` is now an example test not meant to be run in the project itself.
- Cleaned up `package.json` in general.  Updated some old packages.

**Issues:**
- resolves #21
- resolves #22 (makes unnecessary)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "local-home-virual-environment",
+  "name": "local-home-virtual-environment",
```
proppy, line 2: what about `local-home-testing` ?

platform/mock-radio.ts
```diff
@@ -34,10 +34,17 @@ export class MockNetwork {
   ) {
     const key = address + ':' + port.toString();
     if (this.udpListeners.has(key)) {
-      this.udpListeners[key].push(listener);
+      const listeners = this.udpListeners.get(key);
```
proppy, line 37: curious how this is related to packaging? can you add a note in the PR description?

package.json
```diff
@@ -1,5 +1,5 @@
...
   "version": "0.0.1",
   "description": "",
   "private": true,
```
proppy, line 5: remove private?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "local-home-virual-environment",
+  "name": "local-home-virtual-environment",
   "version": "0.0.1",
   "description": "",
```
proppy, line 4: add description?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "local-home-virual-environment",
```
proppy, line 2: add author?

package.json
```diff
@@ -1,5 +1,5 @@
...
+  "name": "local-home-virtual-environment",
   "version": "0.0.1",
```
proppy, line 3: bump/clean dependencies?
- more recent version of ava and typescript?
- use https://github.com/google/gts
- is webpack needed?

package.json
cdaly333: Thanks, was looking for a better name for this.  Will change

platform/mock-radio.ts
cdaly333: Yeah this is not related.  The `tslint` rules in the [sample](https://github.com/actions-on-google/smart-home-local) seem to be more strict than my `eslint` rules.  These null checks were needed to satisfy those rules.  I'll note this in the PR description.

package.json
cdaly333: Webpack not needed.  I'll fix the dependencies and update this PR description to reflect the dependency changes.

package.json
Fleker: I think we should keep as private until we're ready to distribute the artifact.

platform/mock-radio.ts
Fleker: Maybe I don't understand why tslint is throwing an issue. If `this.updListeners.has` resolves to true, then the getter should definitely be true. You can use a non-null assertion `!` at the end of a statement to guarantee it is not-null.

platform/mock-radio.ts
```diff
@@ -34,10 +34,17 @@ export class MockNetwork {
...
+      if (listeners !== undefined) {
+        listeners.push(listener);
+      }
+      const tlisteners = this.udpListeners.get(key);
```
Fleker, line 41: Should not be different from listeners?

platform/mock-radio.ts
```diff
@@ -34,10 +34,17 @@ export class MockNetwork {
...
+      if (tlisteners !== undefined) {
+        console.log(tlisteners.length);
```
Fleker, line 43: Not sure we need to log, but you could write a test to assert the length has changed.

platform/mock-radio.ts
```diff
@@ -34,10 +34,17 @@ export class MockNetwork {
...
+      }
       return;
     }
-    this.udpListeners[key] = [listener];
+    this.udpListeners.set(key, [listener]);
```
Fleker, line 47: Should we overwrite this or push?

platform/mock-radio.ts
cdaly333: It was complaining the value could be undefined.  Maybe you can set a map value as undefined in typescript.  I'll try with `!`

package.json
Fleker: Google LLC

package.json
cdaly333: Done

platform/mock-radio.ts
cdaly333: `!` worked

package.json
cdaly333: Added a simple description.

package.json
cdaly333: Will leave as private for now

package.json
cdaly333: Added `"author": "Google"` to match the Local Home SDK and SDK sample.

package.json
cdaly333: I cleaned the dependencies up and applied `gts`

platform/mock-radio.ts
cdaly333: Left this in from debugging.  Removed.

platform/mock-radio.ts
cdaly333: Left this in from debugging.  Removed.

platform/mock-radio.ts
cdaly333: Should overwrite, but this was missing an early return.  Added the early return.  As a side note I may replace all this in #26

cdaly333:
Removed WIP tag, added description

.prettierrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('gts/.prettierrc.json')
```
Fleker, line 2: Is this required?

package.json
```diff
@@ -1,34 +1,28 @@
 {
-  "name": "local-home-virual-environment",
+  "name": "local-home-testing",
   "version": "0.0.1",
-  "description": "",
+  "description": "Testing environment for Local Fulfillment Apps. Mocks behavior of Local Home Platform.",
+  "author": "Google",
```
Fleker, line 5: Google LLC

package.json
```diff
@@ -1,34 +1,28 @@
...
   "private": true,
+  "main": "./src/index.ts",
```
Fleker, line 7: I believe the main usually points to the compiled Javascript location

src/platform/device-manager.ts
```diff
@@ -4,22 +4,21 @@
...
 
   getProxyInfo(id: string): smarthome.ProxyInfo {
     //TODO(cjdaly) implementation
-    throw new Error("Method not implemented.");
+    throw new Error('Method not implemented.');
   }
 
-  deviceId: string;
-  error: boolean;
+  deviceId: string = '';
```
Fleker, line 20: Do we want the default to be an empty string? Can we leave it null?

src/platform/device-manager.ts
```diff
@@ -4,22 +4,21 @@
...
+  error: boolean = true;
```
Fleker, line 21: We default the error to being true?

test/example/test.ts
```diff
@@ -1,18 +1,19 @@
 /*
- * Tests to verify stub behaviors
+ * Example tests against stubs
```
Fleker, line 2: Are these tests against an external project or testing within itself?

cdaly333: This is intended to be an example test/usage that would be used in an external project.  I'm not sure it belongs in this repository in the long run, and will probably remove it once I resolve #31.

.prettierrc.js
cdaly333: I went ahead and removed it, and updated the local `.prettierrc` with the gts rules.  I also added `prettier` as a dev dependency

package.json
cdaly333: Done

src/platform/device-manager.ts
cdaly333: I added these defaults since tslint was complaining about them never being initialized.  However, since I'm going to implement this full stub in #23, I'm going to remove these variables entirely for now, as well as the `send()` implementation.

package.json
cdaly333: Changed to `./build/src/index.js`

src/platform/device-manager.ts
cdaly333: Deleted this variable. Going to implement in #23.

Fleker:
Looks good, will need to resolve merge conflicts.

cdaly333:
Pulled from master, fixed a few tslint errors and removed the `loadHomeApp` function

src/platform/mock-local-home-platform.ts
```diff
@@ -46,9 +46,9 @@ export class MockLocalHomePlatform implements MockUDPListener {
   private static instance: MockLocalHomePlatform;
 
   private udpScanConfigs: UDPScanConfig[] = [];
-  private mockNetwork: MockNetwork;
-  private deviceManager: smarthome.DeviceManager;
-  private app: AppStub;
+  private mockNetwork!: MockNetwork;
```
Fleker, line 49: Doesn't the `!` operator assert that the variable is not null? Does that make sense in this context?

src/platform/mock-local-home-platform.ts
```diff
@@ -156,8 +156,8 @@ export class MockLocalHomePlatform implements MockUDPListener {
 
     const device = identifyResponse.payload.device;
     console.log('Registering localDeviceId: ' + device.verificationId);
-    this.localDeviceIds.set(device.id, device.verificationId);
-    this.onNewDeviceIdRegistered(device.verificationId);
+    this.localDeviceIds.set(device.id, device.verificationId!);
```
Fleker, line 159: Is `device.verificationId` sometimes undefined?

src/platform/mock-local-home-platform.ts
cdaly333: `tslint` was complaining these value were never initialized and may be undefined, but `mockNetwork` and `app` are set in setters.  I can see how this isn't the best solution, so I just replaced `!` with ` | undefined` types, and null checks for `app` and `mockNetwork` when they are actually used.  For `deviceManager` I added a default value.

src/platform/mock-local-home-platform.ts
cdaly333: I added an early exit if `device.verificationId` is null.  I'll double check this is the expected behavior.

src/platform/mock-local-home-platform.ts
```diff
@@ -129,6 +129,10 @@ export class MockLocalHomePlatform implements MockUDPListener {
 
   // Establish fulfillment path using app code
   async onUDPMessage(msg: Buffer, rinfo: RemoteAddressInfo): Promise<void> {
+    if (this.app === undefined) {
```
Fleker, line 132: Is the right behavior to silently exit, or to indicate an unexpected problem?

cdaly333: Since I'm moving this function to another class in #32, I made the change in #32. (throwing an error and catching it in the test).

src/platform/mock-local-home-platform.ts
cdaly333: See #32

src/platform/smart-home-app.ts
```diff
@@ -3,13 +3,13 @@
...
 
-import { MockLocalHomePlatform } from './mock-local-home-platform';
+import {MockLocalHomePlatform} from './mock-local-home-platform';
 
 export class AppStub implements smarthome.App {
   private version: string;
-  public identifyHandler: smarthome.IntentFlow.IdentifyHandler;
-  public executeHandler: smarthome.IntentFlow.ExecuteHandler;
-  public reachableDevicesHandler: smarthome.IntentFlow.ReachableDevicesHandler;
+  public identifyHandler!: smarthome.IntentFlow.IdentifyHandler;
```
Fleker, line 10: Does this need the `!` syntax?

src/platform/mock-local-home-platform.ts
cdaly333: Since this will be merged first and were a lot of unhandled `undefined` variables, I went ahead and added Error messages.  Will resolve conflicts when merging #32.

src/platform/smart-home-app.ts
cdaly333: Added `| undefined` to these types.  Also went ahead and added null checks which throw errors in the `IdentifyRequest` logic (which will be moved to another class in #32, but I figure is worth adding here just to keep master clean).

package.json
```diff
@@ -1,35 +1,29 @@
...
+  "description": "Testing environment for Local Fulfillment Apps. Mocks behavior of Local Home Platform.",
+  "author": "Google LLC",
   "private": true,
+  "main": "./build/src/index.js",
   "license": "Apache-2.0",
   "scripts": {
     "compile": "tsc",
-    "test": "ava --verbose",
-    "lint": "eslint ."
+    "lint": "tslint --project .",
```
proppy, line 11: shouldn't we be using gts?

cdaly333: I hadn't realized `tslint` was deprecated / not compatible with gts.  I'll switch back to `eslint`. Opened #33

Issue #16, opened 2020-06-24
Implement EXECUTE flow functionality


cdaly333:
Add mock and stub functionality to implement the EXECUTE flow outlined in the design doc.

Issue #20, opened 2020-06-24
Add JSDocs to all functions


cdaly333:

cdaly333:
Closing because JSDocs are being added per-PR

PR #32, opened 2020-07-07
Simplify stubs by removing radio functionality.  Add MockLocalHomePlatform and AppStub tests.

### LINKED ISSUE #27: Simplify Identify functionality by removing scan/broadcast

### LINKED ISSUE #31: Add support for internal tests
- Add typescript support for `ava`
- Add npm script to run an internal test suite.
- Add a basic layout for internal tests

cdaly333:
This PR simplifies the project by stripping away the radio functionality, leaving the Identify flow functionality.  

**Changes:**
- Remove unused `DeviceManagerStub` import
- Remove broadcast functions and scan trigger functions from `MockLocalHomePlatform`
- Remove singleton pattern from `MockLocalHomePlatform` in favor of module-level instance
- Remove `MockUDPListener`, `UDPDevice`, `ScanConfig`, `UDPScanConfig`, `mock-radio.ts`
- Simplified the `listen()` functionality
- Added tests for `MockLocalHomePlatform` and `AppStub`
- Updated project with `ava` typescript support and test script

**Issues:**
- Resolves #27
- Resolves #31 

**Notes:**
- `mock-radio.ts` was removed since it isn't being used currently, but I'm anticipating adding it back when resolving #16 and #23


platform/mock-radio.ts
```diff
@@ -58,16 +58,43 @@ export class MockNetwork {
...
-    this.udpMessageAction = messageAction;
+  public getDeviceId(): string {
+    return this.deviceId;
+  }
+
+  public getDevicePort(): number {
+    return this.port;
+  }
+
+  public triggerIdentify(
```
Fleker, line 87: This triggers an IDENTIFY intent? Could be helpful to include some JSDoc.

platform/mock-radio.ts
```diff
@@ -58,16 +58,43 @@ export class MockNetwork {
...
+    discoveryBuffer: Buffer,
+    discoveryPort: number,
+    disvoceryAddress: string
```
Fleker, line 90: typo, "discovery"

test/test.ts
```diff
@@ -17,79 +13,58 @@ import {
 
 // Tests a UDP identify flow end-to-end
 test('udp-device-connects', async (t) => {
+  // Common address for all UDP messaging
+  const updAddress: string = '255.255.255.255';
```
Fleker, line 17: If this can be a common variable, can it be made a global variable in this file?

test/test.ts
```diff
@@ -17,79 +13,58 @@ import {
...
+  const mockDevice = new UDPDevice(
+    'test-device-id',
+    mockNetwork,
+    12345,
+    updAddress
   );
 
-  loadHomeApp('../home-app/bundle');
+  // Run Home App and set handlers
+  // loadHomeApp('../bundle');
```
Fleker, line 43: Do we no longer need this? If it's commented out, it'd be better to just delete it.

platform/mock-radio.ts
```diff
@@ -58,16 +58,43 @@ export class MockNetwork {
...
+  ): void {
+    this.network.sendUDPMessage(
+      discoveryBuffer,
+      discoveryPort,
+      disvoceryAddress,
```
ToniCorinne, line 95: same typo

platform/mock-radio.ts
cdaly333: Good point this isn't directly clear.  I renamed it to `sendDiscoveryBuffer` since that's more accurate.  The `MockLocaIHomePlatform` is what receives the `discoveryPlatform` and triggers the IDENTIFY request.  Also added a JSDoc.  I'm planning on adding JSDocs later, but I'll start adding JSDocs here and there as to not do it all at once (see #20).

platform/mock-radio.ts
cdaly333: Replaced with `UDPScanConfig` since that's where the port and address originate from anyways

test/test.ts
cdaly333: Done.

test/test.ts
cdaly333: Done.

platform/mock-radio.ts
cdaly333: Removed.

test/test.ts
```diff
@@ -3,93 +3,67 @@
...
+const updAddress: string = '255.255.255.255';
+
 // Tests a UDP identify flow end-to-end
 test('udp-device-connects', async (t) => {
   // First, create a scan configuration
-  const scanConfig = new UDPScanConfig(
+  const scanConfig: UDPScanConfig = new UDPScanConfig(
     ScanState.Unprovisioned,
-    '255.255.255.255',
+    updAddress,
```
proppy, line 22: shouldn't we also get rid of this, since we don't send the broadcast anymore?

platform/mock-local-home-platform.ts
```diff
@@ -159,15 +146,4 @@ export class MockLocalHomePlatform implements MockUDPListener {
     this.localDeviceIds.set(device.id, device.verificationId);
     this.onNewDeviceIdRegistered(device.verificationId);
   }
-
-  public addUDPScanConfig(scanConfig: UDPScanConfig) {
-    this.udpScanConfigs.push(scanConfig);
```
proppy, line 164: do we still need the `udpScanConfigs` field and the `ScanConfig` type?

platform/mock-radio.ts
```diff
@@ -58,16 +60,47 @@ export class MockNetwork {
...
+  public getDevicePort(): number {
+    return this.port;
+  }
+
+  /**
+   * Sends a `discoveryBuffer` to the port specified in a `UDPScanConfig`
+   * @param discoveryBuffer
+   * @param scanConfig
+   */
+  public sendDiscoveryBuffer(
```
proppy, line 94: Should we move this to the platform class and remove the scanConfig parameter? It's not really strictly tied to the radio simulation anymore.

test/test.ts
```diff
@@ -3,93 +3,67 @@
...
 
+  // Assert listen() was called and handlers were loaded
   t.is(mockLocalHomePlatform.isHomeAppReady(), true);
 
-  const connectedDeviceId = mockLocalHomePlatform.getNextDeviceIdRegistered();
-
-  // Start scanning
-  mockLocalHomePlatform.triggerScan();
+  // Create promise to catch next deviceId registered
+  const connectedDeviceId: Promise<string> = mockLocalHomePlatform.getNextDeviceIdRegistered();
```
proppy, line 49: We should be able to move this after the sendDiscoveryBuffer.

test/test.ts
cdaly333: Yeah we can.  Removed.

platform/mock-local-home-platform.ts
cdaly333: Removed.

test/test.ts
cdaly333: Moved inline with the test.

platform/mock-radio.ts
cdaly333: Yeah, that makes sense.  I also removed the reference to the `MockNetwork` and the `MockUDPListener` implementation for now since it's now not being used.  I anticipate needing it later for `EXECUTE` #16.

cdaly333:
Removed `ScanConfig` and `UDPScanConfig`

platform/mock-local-home-platform.ts
```diff
@@ -97,27 +45,26 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
+      });
     });
   }
 
-  // Establish fulfillment path using app code
-  async onUDPMessage(msg: Buffer, rinfo: RemoteAddressInfo): Promise<void> {
-    console.log('received discovery payload:', msg, 'from:', rinfo);
+  /**
+   * Takes a `discoveryBuffer` and passes it to the fulfillment app in an `IdentifyRequest`
+   * @param discoveryBuffer
```
Fleker, line 64: Typically a JSDoc parameter includes a short description of the parameter.

cdaly333: Done.

cdaly333:
Tested example test with [LHSDK Sample](https://github.com/actions-on-google/smart-home-local)

src/platform/mock-local-home-platform.ts
```diff
@@ -66,9 +68,12 @@ export class MockLocalHomePlatform {
   public async triggerIdentify(discoveryBuffer: Buffer): Promise<void> {
     console.log('received discovery payload:', discoveryBuffer);
 
-    if (this.app == null) {
+    if (this.app === undefined) {
+      throw new Error('Cannot trigger IdentifyRequest: App was undefined');
+    }
+    if (!this.isHomeAppReady()) {
```
Fleker: Are you going to write tests to check these errors are thrown correctly?

cdaly333: Yeah, I plan to add tests soon ( Re: #31), unless you think I should just go ahead and do it here.

Fleker: Generally you add tests at the same time as behavior to ensure your behavior works as expected.

cdaly333: Will add

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,91 @@
...
+  ERROR_UNDEFINED_VERIFICATIONID,
+} from '../../src/platform/mock-local-home-platform';
+import {AppStub} from '../../src/platform/smart-home-app';
+import {injectSmarthomeStubs} from '../../src';
+
+test.before(t => {
+  injectSmarthomeStubs();
+});
+
+test('only-trigger-identify-when-platform-initialized', async t => {
```
cdaly333, line 17: Would have broken into further tests but singleton pattern makes this difficult.  Any suggestions?

src/platform/mock-local-home-platform.ts
cdaly333: Done

test/platform/test-platform-identify.ts
Fleker: If you're having issues with the singleton, you may want to come up with a way to mock the instance or have a way to do a reset. Being able to initialize to a clean state will be useful.

cdaly333: Yeah good point - this is a testing tool anyways.  I'll add a method to reset the state

cdaly333: Added optional parameter to reset state in singleton getter.  Split up tests.

cdaly333:
Added tests for `triggerIdentify()`

src/platform/mock-local-home-platform.ts
```diff
@@ -40,8 +49,10 @@ export class MockLocalHomePlatform {
   }
 
   //  Singleton getter
-  public static getInstance(): MockLocalHomePlatform {
-    if (!MockLocalHomePlatform.instance) {
+  public static getInstance(
+    resetState: boolean = false
```
Fleker, line 53: Do you think we'll ever want to include more parameters at some point to this getter? Should we include some JSDoc for this method?

cdaly333: I'm not sure I want to allow other parameters in `getInstance()` since that complicates the assumptions classes could make when obtaining the singleton instance through this getter.  The reset state parameter should only be used externally when setting up a test, and not internally in the `MockLocalHomePlatform` or `AppStub`, for example.  Will add a JSDoc.

src/platform/mock-local-home-platform.ts
```diff
@@ -106,40 +48,47 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
-  private sendUDPBroadcast(scanConfig: UDPScanConfig) {
-    const packetBuffer = Buffer.from(scanConfig.discoveryPacket, 'hex');
-    this.mockNetwork!.sendUDPMessage(
-      packetBuffer,
-      scanConfig.broadcastPort,
-      scanConfig.broadcastAddress,
-      scanConfig.listenPort,
-      scanConfig.broadcastAddress
-    );
+  public async getNextDeviceIdRegistered(): Promise<string> {
```
Fleker, line 70: Can you add JSDoc for this method?

src/platform/mock-radio.ts
```diff
@@ -27,7 +27,7 @@ export class MockNetwork {
     this.udpListeners = new Map<string, MockUDPListener[]>();
   }
 
-  registerUDPListener(
+  public registerUDPListener(
```
Fleker, line 30: Can you add JSDoc for this method?

src/platform/mock-radio.ts
```diff
@@ -64,19 +64,29 @@ export class MockNetwork {
...
-  ) => void = () => {};
+export class UDPDevice {
+  private deviceId: string;
+  private port: number;
+  private address: string;
+  private network: MockNetwork;
 
-  onUDPMessage(msg: Buffer, rinfo: RemoteAddressInfo): void {
-    this.udpMessageAction(msg, rinfo);
+  public constructor(
```
Fleker, line 73: Can you add JSDoc here?

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,131 @@
...
+test.before(t => {
+  injectSmarthomeStubs();
+});
+
+test('trigger-identify-with-undefined-app-throws', async t => {
+  const mockLocalHomePlatform = MockLocalHomePlatform.getInstance(true);
+  t.is(
+    (
+      await t.throwsAsync<Error>(
+        mockLocalHomePlatform.triggerIdentify(DISCOVERY_BUFFER)
```
Fleker, line 26: Can you explain why this may be an error?

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+});
+
+test('trigger-identify-without-listen-throws', async t => {
+  const mockLocalHomePlatform = MockLocalHomePlatform.getInstance(true);
+  const app: AppStub = new AppStub(APP_VERSION);
+  t.is(
+    (
+      await t.throwsAsync<Error>(
+        mockLocalHomePlatform.triggerIdentify(DISCOVERY_BUFFER)
```
Fleker, line 39: Can you explain why this may be an error?

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+});
+
+test('trigger-identify-with-undefined-handler-throws', async t => {
+  const mockLocalHomePlatform = MockLocalHomePlatform.getInstance(true);
+  const app: AppStub = new AppStub(APP_VERSION);
+  app.listen();
+  t.is(
+    (
+      await t.throwsAsync<Error>(
```
Fleker, line 52: Can you explain why this may be an error?

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,131 @@
...
+          id: DEVICE_ID,
+        },
+      },
+    };
+  };
+  app.onIdentify(invalidIdentifyHandler).listen();
+  t.is(
+    (
+      await t.throwsAsync<Error>(
+        mockLocalHomePlatform.triggerIdentify(DISCOVERY_BUFFER)
```
Fleker, line 78: Can you explain why this may be an error?

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,131 @@
...
+  t.is(
+    mockLocalHomePlatform.getLocalDeviceIdMap().values().next().value,
+    localDeviceId
+  );
+});
+
+test('trigger-identify-without-reset-state-throws', async t => {
+  const oldPlatform = MockLocalHomePlatform.getInstance(true);
+  const app: AppStub = new AppStub(APP_VERSION);
+  const mockLocalHomePlatform = MockLocalHomePlatform.getInstance(false);
```
Fleker, line 122: Do we need to get this twice?

src/platform/mock-local-home-platform.ts
cdaly333: Done

src/platform/mock-radio.ts
cdaly333: Done

src/platform/mock-radio.ts
cdaly333: Done

test/platform/test-platform-identify.ts
cdaly333: Added comments explaining all tests.  This test throws because there was no `App` ever created and attached to the Local Home Platform.  There would be no handler to pass a buffer to.

test/platform/test-platform-identify.ts
cdaly333: Added comments explaining this.  `listen()` indicates all handlers have been set by setting a flag.  `MockLocalHomePlatform` returns this error when the flag has not been set prior to calling the handler.

test/platform/test-platform-identify.ts
cdaly333: Added comments.  When `listen()` has been called but there was never an `identifyHandler` passed into `AppStub`, the `MockLocalHomePlatform` returns this error.

test/platform/test-platform-identify.ts
cdaly333: Added comments.  When there is an `undefined` `verificationId` is returned from the `identifyHandler`, the Platform can't associate a local device id and returns an error.

test/platform/test-platform-identify.ts
cdaly333: This tests the behavior of the `getInstance()` reset state flag.  The first `getInstance()` call is necessary so the difference between the two flag values can be compared.  I replaced this test with the opposite case since this failing case is kind of arbitrary.  The new test initializes a `MockLocalHomePlatform`, sets an `App`, then creates a new `MockLocalHomePlatform` with `resetState = true`, and assets that there is no `App` set in the new `MockLocalHomePlatform`

src/platform/mock-local-home-platform.ts
```diff
@@ -67,6 +58,10 @@ export class MockLocalHomePlatform {
     });
   }
 
+  /**
+   * Asyncronously returns the next localDeviceId registered to the Local Home Platform.
+   * This localDeviceId is referred to as the verificationId in the IdentifyResponse
+   */
```
Fleker, line 64: Should add an `@returns`  tag

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,91 @@
...
   }
 
-  //  Singleton getter
-  public static getInstance(): MockLocalHomePlatform {
-    if (!MockLocalHomePlatform.instance) {
+  /**
+   * Returns the static singleton instance, creating it if needed.
+   * @param resetState  Whether or not to force a new instance.  Useful for initializing tests.
+   */
+  public static getInstance(
```
proppy, line 46: Curious if this need to be a singleton?

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,91 @@
...
-        udpScanConfig.listenPort,
-        udpScanConfig.broadcastAddress
-      );
+  /**
+   * Asyncronously returns the next localDeviceId registered to the Local Home Platform.
+   * This localDeviceId is referred to as the verificationId in the IdentifyResponse
+   */
+  public async getNextDeviceIdRegistered(): Promise<string> {
+    return new Promise(resolve => {
+      this.newDeviceRegisteredActions.push(localDeviceId => {
```
proppy, line 67: can't simply push resolve?

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,91 @@
...
   }
 
-  public getDeviceManager(): smarthome.DeviceManager {
-    return this.deviceManager;
-  }
-
-  public getLocalDeviceIdMap(): Map<string, string> {
-    return this.localDeviceIds;
+  private onNewDeviceIdRegistered(localDeviceId: string) {
+    this.newDeviceRegisteredActions.forEach(newDeviceRegisteredAction => {
```
proppy, line 56: Does this need to be a list? (do we are of waiting for more than one registration at a time)

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,91 @@
...
-      scanConfig.listenPort,
-      scanConfig.broadcastAddress
-    );
-  }
+  /**
+   * Takes a `discoveryBuffer` and passes it to the fulfillment app in an `IdentifyRequest`
+   * @param discoveryBuffer  the buffer to be included in the `IdentifyRequest` scan data
+   */
+  public async triggerIdentify(discoveryBuffer: Buffer): Promise<void> {
+    console.log('Received discovery payload:', discoveryBuffer);
```
proppy, line 78: Do we want to keep the logging? maybe set it as `.debug`?

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,91 @@
...
-  async onUDPMessage(msg: Buffer, rinfo: RemoteAddressInfo): Promise<void> {
+    // Need a reference to the `App` instance to be able to pass an `IdentifyRequest`
     if (this.app === undefined) {
-      throw new Error('Cannot trigger IdentifyRequest: App was undefined');
-    }
-    if (!this.isHomeAppReady()) {
-      throw new Error(
-        'Cannot trigger IdentifyRequest: listen() was not called'
-      );
+      throw new Error(ERROR_UNDEFINED_APP);
```
proppy, line 82: Would we make it so, `this.app` is never undefined? (maybe by making it required to construct the platform?)

src/platform/smart-home-app.ts
```diff
@@ -12,19 +12,24 @@ export class AppStub implements smarthome.App {
...
+    return this.allHandlersSet;
+  }
+
   getDeviceManager(): smarthome.DeviceManager {
     return MockLocalHomePlatform.getInstance().getDeviceManager();
   }
 
   listen(): Promise<void> {
-    MockLocalHomePlatform.getInstance().notifyHomeAppReady();
+    this.allHandlersSet = true;
```
proppy, line 32: Does it actually guarantee that all the handlers have been set, should we test for it?

src/platform/mock-radio.ts
```diff
@@ -3,6 +3,11 @@
  */
 
 export interface MockUDPListener {
```
proppy, line 5: do we need to keep this? given that it's not being used anymore?

test/example/test.ts
```diff
@@ -5,93 +5,57 @@
...
-      scanConfig.broadcastAddress
-    );
-  });
-
-  //Mock network needs to listen for device response
-  mockNetwork.registerUDPListener(
-    mockDevice,
-    scanConfig.broadcastPort,
-    scanConfig.broadcastAddress
+  const mockDevice = new UDPDevice(
```
proppy, line 23: is the mockDevice still needed? you could simply pass the deviceId and devicePort as parameter for trigger identify.

src/platform/mock-local-home-platform.ts
cdaly333: It's a singleton so an arbitrary `AppStub` instance can access the single instance of `MockLocalHomePlatform`.  This instance also needs to be accessed in tests in order to trigger intents and assert against.  Another option I was considering is to have a static event that fires when an `AppStub` is created.  Then a `MockLocalHomePlatform` could listen for that event and create the association.  That seems functionally the same as the current implementation, but would get rid of the "reset instance" flag.

The other option is to have the user re-implement their app (`App` constructor, setting handlers, call to `listen()`) within their test scope, so they can inject the dependency in that scope.

Open to suggestions.

src/platform/mock-local-home-platform.ts
cdaly333: Could you show me what you mean?  Currently using this constructor so the action has access to `resolve`

src/platform/mock-local-home-platform.ts
cdaly333: I'm not sure how I would get access to the created `App`.  In testing this would catch the case where the developer doesn't create an `App` class at all.

src/platform/mock-local-home-platform.ts
cdaly333: Done.

src/platform/mock-local-home-platform.ts
cdaly333: I was able to remove this function and the action entirely by just returning the local device id in the `triggerIdentify` call

src/platform/mock-local-home-platform.ts
cdaly333: Done

src/platform/smart-home-app.ts
cdaly333: Added checks and errors for this

src/platform/mock-local-home-platform.ts
cdaly333: Was able to remove this pattern and just return the local device id in `triggerIdentify`

src/platform/mock-radio.ts
cdaly333: Removed `mock-radio.ts` for this PR since nothing is being used yet.  I anticipate adding it back for #23 and #16

test/example/test.ts
cdaly333: Removed `mock-radio.ts` for this PR since nothing is being used yet.  I anticipate adding it back for #23 and #16

cdaly333:
I updated the PR name and description since this PR is swaying a bit from the original description.  Notable changes:

- I removed the example test and added in tests for `AppStub` and `MockLocalHomePlatform`.  The new  internal tests illustrate and assert the same functionality compared to the example test.
- I removed all radio functionality for this PR since it wasn't being used.
- Added comments and JSDocs for public functions and tests

src/platform/mock-local-home-platform.ts
proppy: > It's a singleton so an arbitrary AppStub instance can access the single instance of MockLocalHomePlatform

Something to consider is if you'd ever need to associate multiple applications to a single platform stub instance.
If that's not the case maybe you could simply create a throwaway platform stub along side the app and keep a cross reference?

> This instance also needs to be accessed in tests in order to trigger intents and assert against. 

Could the tests import and explicit create a MockPlatform instance per test? That would also dispense you from implementing any kind of reset mechanism.

> The non-static option is to have the user re-implement their App creation (App constructor, setting handlers, call to listen()) within their test scope, so they can inject the dependency in that scope.

Another element is how explicit you want to make the platform stubbing from tests, one way to surface it would be that once imported/initialized `smarthome.App` is exposed in lieu of `AppStub` just like if you imported the real thing.

cdaly333: > Something to consider is if you'd ever need to associate multiple applications to a single platform stub instance.
If that's not the case maybe you could simply create a throwaway platform stub along side the app and keep a cross reference?

There should only ever be one instance of the `App` and one instance of the `MockLocalHomePlatform`.  However, the intent is that `App` gets created in the developer's default code (for example, [this file](https://github.com/actions-on-google/smart-home-local/blob/master/app/index.ts)), while the `MockLocalHomePlatform` gets created in the testing scope.  The static instance is the cross-referencing mechanism.  If I created the `MockLocalHomePlatform` within the `App` stub, I couldn't access either from a test (without some other static mechanism)

> Could the tests import and explicit create a MockPlatform instance per test? That would also dispense you from implementing any kind of reset mechanism.

Yes, but I would need some way of notifying the next `App` stub of this instance.

> Another element is how explicit you want to make the platform stubbing from tests, one way to surface it would be that once imported/initialized smarthome.App is exposed in lieu of AppStub just like if you imported the real thing.

Good point, and something I'll change in this PR.

I think my takeaway is that it may be worth ditching the idea of having the user be able to run their code without modification, with all the linking and cross-referencing happening automatically in the background.  Instead, requiring the user to explicitly import and construct an app might lead to a more flexible testing environment, and imply less "hacky" workarounds like a singleton.  Will keep looking into this.

proppy: > However, the intent is that App gets created in the developer's default code (for example, this file), while the MockLocalHomePlatform gets created in the testing scope. The static instance is the cross-referencing mechanism.

If `smarthome.App` (the stub version) create its own platform, there shouldn't be a need for static cross-reference mecanism (the platform constructor could take the `AppStub` as a parameter and the `AppStub` could ref it as a member after creating it).

In order to trigger intent from test, you could have top-level module function (or `AppStub` methods) that downcast the `smarthome.App` instance in order to get back the platform.

> Yes, but I would need some way of notifying the next App stub of this instance.

Could the next `App` create their own platform?

> I think my takeaway is that it may be worth ditching the idea of having the user be able to run their code without modification, with all the linking and cross-referencing happening automatically in the background. Instead, requiring the user to explicitly import and construct an app might lead to a more flexible testing environment, and imply less "hacky" workarounds like a singleton. Will be mindful of this if it grows into a larger issue.

That's also a good way to think about it, you could always layer the magic back later in this prove too tedious.

cdaly333: I pushed a implementation of a top-level module function rather than a singleton.  Closing this thread.

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,124 @@
...
+  });
+});
+
+/**
+ * Tests that `listen()` was called on the created App.
+ * This is a required flag that indicates handlers have been set.
+ */
+test('trigger-identify-without-listen-throws', async t => {
+  const app: smarthome.App = new smarthome.App(APP_VERSION);
+  const mockLocalHomePlatform = getActiveAppStub().getLocalHomePlatform();
```
proppy, line 57: you don't really need `getActiveAppStub` since you already have the `app` in this context.

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,124 @@
...
+    message: ERROR_LISTEN_NOT_CALLED,
+  });
+});
+
+/**
+ * Tests that the returned `IdentifyResponse` contains a verificationId.
+ */
+test('trigger-identify-with-undefined-verificationId-throws', async t => {
+  const app: smarthome.App = new smarthome.App(APP_VERSION);
+  const mockLocalHomePlatform = getActiveAppStub().getLocalHomePlatform();
```
proppy, line 69: you don't really need `getActiveAppStub` since you already have the `app` in this context.

test/platform/test-platform-identify.ts
```diff
@@ -0,0 +1,124 @@
...
+        device: {
+          id: DEVICE_ID,
+          verificationId: localDeviceId,
+        },
+      },
+    };
+  };
+  const app: smarthome.App = new smarthome.App(APP_VERSION);
+  app.onIdentify(validIdentifyHandler).onExecute(EXECUTE_HANDLER).listen();
+  const mockLocalHomePlatform = getActiveAppStub().getLocalHomePlatform();
```
proppy, line 108: you don't really need `getActiveAppStub` since you already have the `app` in this context.

src/platform/stub-setup.ts
```diff
@@ -16,6 +19,23 @@ const smarthomeStub: {
...
   (global as any).smarthome = smarthomeStub;
-}
\ No newline at end of file
+}
+
+// Module-level variable to hold the active instance the AppStub
+var activeAppStub: AppStub | undefined;
+
+// Module-level getter for the active AppStub
+export function getActiveAppStub(): AppStub {
```
proppy, line 31: Sorry for lack of details with my earlier comments, when I mentioned `top-level module function`, I meant function that would take a `smarthome.App` as parameter and promote it to a `smarthome.AppStub` to get the platform out of it and dispatch an intent.

I don't believe you need to get keep any global state, as the test already know about the app they need to test.

test/example/test.ts
```diff
@@ -1,97 +0,0 @@
-/*
```
proppy, line 1: It would be nice to keep this (at least the modified version of it) as a canonical example on how the library should be used.

test/platform/test-platform-identify.ts
cdaly333: Done

test/platform/test-platform-identify.ts
cdaly333: Done.

test/platform/test-platform-identify.ts
cdaly333: Done

src/platform/stub-setup.ts
cdaly333: Oh I see.  Done. The only thing is now the developer will need to re-implement the creation of their App in the same context as their test (as opposed to being able to use the App creation code in their entry point).  However, like you said I think it's something that can be layered later.

src/platform/stub-setup.ts
```diff
@@ -4,6 +4,9 @@
  */
```
cdaly333, line 4: Fixing this file

cdaly333: Done

test/example/test.ts
cdaly333: Done

src/platform/stub-setup.ts
proppy: Wouldn't the app creation automatically use the stub as long as install the `smarthome` namespace before loading the app?

src/platform/mock-local-home-platform.ts
```diff
@@ -1,101 +1,32 @@
...
- * - Interacts with bundled HomeApp
- * - Accepts scan config
- * - Interacts with virtual network
  */
 
 /// <reference types="@google/local-home-sdk" />
-import {MockNetwork, MockUDPListener, RemoteAddressInfo} from './mock-radio';
-import {DeviceManagerStub} from './device-manager';
 import {AppStub} from './smart-home-app';
+import {DeviceManagerStub} from './device-manager';
```
proppy, line 7: Seems that only bit of functionality left in the `MockPlatform` is the intent dispatching, would it make sense to fold it into `AppStub`?

src/platform/stub-setup.ts
```diff
@@ -16,6 +17,17 @@ const smarthomeStub: {
...
-export function injectSmarthomeStubs() {
+// Injects stubs into the global scope, allowing implicit usage for testing
+export function injectSmarthomeStubs(): void {
   (global as any).smarthome = smarthomeStub;
-}
\ No newline at end of file
+}
+
+// Promotes App to AppStub
+export function extractMockLocalHomePlatform(
```
proppy, line 26: I wonder if we actually need this, it should be always safe for developer to downcast to `AppStub` within their test.

src/platform/stub-setup.ts
```diff
@@ -16,6 +17,17 @@ const smarthomeStub: {
   },
 };
 
-export function injectSmarthomeStubs() {
+// Injects stubs into the global scope, allowing implicit usage for testing
+export function injectSmarthomeStubs(): void {
```
proppy, line 21: should we do this implicitly on the main module import?

src/platform/mock-local-home-platform.ts
```diff
@@ -106,39 +37,18 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
+  /**
+   * Takes a `discoveryBuffer` and passes it to the fulfillment app in an `IdentifyRequest`
+   * @param discoveryBuffer  The buffer to be included in the `IdentifyRequest` scan data
+   * @returns  The next localDeviceId registered to the Local Home Platform
+   */
+  public async triggerIdentify(discoveryBuffer: Buffer): Promise<string> {
+    console.debug('Received discovery payload:', discoveryBuffer);
+
+    // Cannot start processing until all handlers have been set on the `App`
+    if (!this.app.areAllHandlersSet()) {
```
proppy, line 49: couldn't we simply test the identifyHandler directly?

src/platform/mock-local-home-platform.ts
cdaly333: It's worth considering, but I'm not sure I want to combine them at this point.  (I also think it paints a component diagram of the system being mocked, i.e. a developer reading the code conceptualizes the idea of the developer's app talking to a "platform" layer.)

src/platform/stub-setup.ts
cdaly333: I think this function is okay to keep around for now.  Since the developer would need the `MockLocalHomePlatform` for essentially every test, I think wrapping this cast would save some repetition in the long run.

src/platform/mock-local-home-platform.ts
cdaly333: We could, but this also imposes that `listen()` was called, and that the Execute handler was also set.  I don't know if we want to allow the Identify handler to be set without an Execute handler, since I'm assuming that would be an invalid state for a local fulfillment app.

src/platform/stub-setup.ts
cdaly333: Yeah, that makes more sense.  Done

src/platform/mock-local-home-platform.ts
proppy: I guess what I find confusing is that the flag is called 'handlerSet', while the error message is about `listen not being called'.

Maybe we could move that get to the platform accessor? You shouldn't be able to get the platform and exercise test functions if `listen` has not be called.

And keep a handler specific check here?

src/platform/stub-setup.ts
proppy: What about calling this `withMockPlatform(app)` and updating the public interface to be fluent?

src/platform/mock-local-home-platform.ts
```diff
@@ -106,39 +37,18 @@ export class MockLocalHomePlatform implements MockUDPListener {
...
-  // Establish fulfillment path using app code
-  async onUDPMessage(msg: Buffer, rinfo: RemoteAddressInfo): Promise<void> {
-    if (this.app === undefined) {
-      throw new Error('Cannot trigger IdentifyRequest: App was undefined');
+  /**
+   * Takes a `discoveryBuffer` and passes it to the fulfillment app in an `IdentifyRequest`
+   * @param discoveryBuffer  The buffer to be included in the `IdentifyRequest` scan data
+   * @returns  The next localDeviceId registered to the Local Home Platform
+   */
+  public async triggerIdentify(discoveryBuffer: Buffer): Promise<string> {
```
proppy, line 45: curious if we could return `this` here to allow chaining it with assertion? (see https://github.com/kozhevnikov/proxymise for a way to enable fluent async interface).

src/platform/mock-local-home-platform.ts
```diff
@@ -106,39 +37,18 @@ export class MockLocalHomePlatform implements MockUDPListener {
     return this.localDeviceIds;
```
proppy, line 37: Do we care about surfacing the whole mapping? or should we surface assertion helper instead ex: `isDeviceRegistered()` ?

src/platform/mock-local-home-platform.ts
cdaly333: I moved the assertion to the `App` side and added a handler-specific check.

src/platform/mock-local-home-platform.ts
cdaly333: Looks interesting, I opened an issue for it (#35)

src/platform/mock-local-home-platform.ts
cdaly333: Done.

src/platform/mock-local-home-platform.ts
```diff
@@ -1,144 +1,59 @@
...
+   * Takes a `discoveryBuffer` and passes it to the fulfillment app in an `IdentifyRequest`
+   * @param discoveryBuffer  The buffer to be included in the `IdentifyRequest` scan data
+   * @returns  The next localDeviceId registered to the Local Home Platform
+   */
+  public async triggerIdentify(discoveryBuffer: Buffer): Promise<string> {
+    console.debug('Received discovery payload:', discoveryBuffer);
+
+    // Cannot start processing until all handlers have been set on the `App`
+    if (!this.app.identifyHandler) {
+      return Promise.reject(new Error(ERROR_UNDEFINED_IDENTIFY_HANDLER));
```
Fleker, line 55: You don't need to wrap values in rejects or resolves. Running `await` on any function that doesn't return a Promise will work fine. `await (1+1)` returns 2 without having to be explicit about it being a promise.

So here you can just `throw new Error(ERROR_UNDEFINED_IDENTITY_HANDLER)`

and you can just `return device.verificationId` without wrapping.

src/platform/smart-home-app.ts
```diff
@@ -5,26 +5,55 @@
...
   getDeviceManager(): smarthome.DeviceManager {
-    return MockLocalHomePlatform.getInstance().getDeviceManager();
+    return this.mockLocalHomePlatform.getDeviceManager();
   }
 
   listen(): Promise<void> {
-    MockLocalHomePlatform.getInstance().notifyHomeAppReady();
+    if (this.identifyHandler === undefined) {
+      this.allHandlersSet = false;
+      return Promise.reject(new Error(ERROR_LISTEN_WITHOUT_IDENTIFY_HANDLER));
```
Fleker, line 50: Same here, you do not need to wrap throwing an error.

src/platform/mock-local-home-platform.ts
cdaly333: Done and noted.

src/platform/smart-home-app.ts
cdaly333: Done

src/platform/smart-home-app.ts
```diff
@@ -5,26 +5,55 @@
...
 
   getDeviceManager(): smarthome.DeviceManager {
-    return MockLocalHomePlatform.getInstance().getDeviceManager();
+    return this.mockLocalHomePlatform.getDeviceManager();
   }
 
   listen(): Promise<void> {
-    MockLocalHomePlatform.getInstance().notifyHomeAppReady();
+    if (this.identifyHandler === undefined) {
+      this.allHandlersSet = false;
```
proppy, line 49: nit: do we need to set this, since this is the default?

cdaly333: Removed.

PR #39, opened 2020-07-29
Update JSDocs, linter and dependencies

cdaly333:
This PR is mostly stylistic changes.  Formatted a lot of comments that were excessively long and switched back to `eslint` in compliance with `gts`.  I also updated UDP response spec/functionality to comply with LHSDK 1.1.0. 

**Changes:**
- Update to `eslint` from `tslint`
- Fix `eslint` rule violations
- Minor tweaks to `package.json`
- Fix some comment punctuation and phrasing (add periods to all comments)
- Shorten comments to more closely adhere to 80 character line limit
- Update for compatibility with LHSDK 1.1.0

**Issues:**
- resolves #33 
- resolves #40

PR #41, opened 2020-07-29
Add a basic command line intent-firing interface

### LINKED ISSUE #3: Implement CLI for manual intent dispatch

cdaly333:
This PR adds a basic command line interface that can fire Identify and Execute intents.  

The command `npm run cli -- --app_path APP_PATH` can be used to start the CLI, where `APP_PATH` is a JavaScript file that creates an instance of a Local Fulfillment App.  There is validation to make sure the referenced file correctly initializes a Local Fulfillment App.  

The constructor of `smarthome.App` is overridden to reference the created `AppStub` within a worker thread. A `CommandProcessor` on the main thread parses user input, does some validation, then sends an object to a worker thread which triggers the corresponding Intent in the `MockLocalHomePlatform` instance.

**Changes**
- Add `src/cli/app_instance.ts` `src/cli/command-processor.ts`, `src/cli/main.ts`, `src/cli/README.md`
- Modify `package.json` with new cli script

**Notes:**
- Will add automatic tests and more examples in the `README`
- Currently the CLI is able to send Identify and Execute intents without mocking `DeviceManager` responses, so the functionality is limited.  Will explore further, but I'm not sure if it's worth fleshing this out too far since for any "coordinated" testing, the automated testing is far superior.  One thing I may add, for completion, is a CLI prompt when `DeviceManager` commands are intercepted.

**Issues:**
- Resolves #3 

Depends on #39


package.json
```diff
@@ -8,22 +8,38 @@
   "license": "Apache-2.0",
   "scripts": {
     "compile": "tsc",
-    "lint": "tslint --project .",
     "test": "ava",
     "clean": "gts clean",
-    "fix": "gts fix"
+    "cli": "node ./build/src/cli/main.js",
+    "lint": "eslint './src/**/*.ts' './test/**/*.ts'"
```
Fleker, line 14: Could you lint '**/*.ts' to catch all Typescript files?

package.json
```diff
@@ -8,22 +8,38 @@
...
+  },
+  "dependencies": {
+    "@types/yargs": "^15.0.5",
+    "ava": "^3.10.0",
+    "yargs": "^15.4.1"
   },
   "devDependencies": {
     "@ava/typescript": "^1.1.1",
-    "@google/local-home-sdk": "^1.0.0",
+    "@google/local-home-sdk": "^1.1.0",
```
Fleker, line 23: Can we update this dependency in a separate pull request just to simplify things?

src/cli/app-instance.ts
```diff
@@ -0,0 +1,116 @@
...
+smarthomeStub.App = CliAppStub;
+(global as any).smarthome = smarthomeStub;
+
+/**
+ * Validate the filepath.
+ */
+try {
+  // Strips the file extension off the filepath,
+  const modulePath = workerData.substring(0, workerData.lastIndexOf('.'));
+  fs.existsSync(path.relative('./', modulePath));
```
Fleker, line 35: Doesn't this return true/false rather than throw?

Fleker: https://nodejs.org/api/fs.html#fs_fs_existssync_path

src/cli/main.ts
```diff
@@ -0,0 +1,44 @@
+import yargs from 'yargs/yargs';
```
Fleker, line 1: Wouldn't this file rather be called `index.ts`?

src/cli/main.ts
```diff
@@ -0,0 +1,44 @@
...
+import {Worker} from 'worker_threads';
+import {CommandProcessor} from './command-processor';
+
+const APP_INSTANCE_PATH = './build/src/cli/app-instance.js';
+
+/**
+ * Process initial command-line arguments.
+ */
+const argv = yargs()
+  .usage('Usage $0 --app_path PATH')
```
Fleker, line 11: Can we simplify the command-line arg to a single word to simplify the call? ie. `--app /path/to/app`.

src/cli/main.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  worker.on('error', error => {
+    console.log(
+      'An error occured while trying to initialize the command line interface:\n' +
+        error.toString()
+    );
+  });
+
+  worker.on('message', async message => {
+    if (message === 'ready') {
```
Fleker, line 35: Is 'ready' a known constant based on an enum?

src/cli/main.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Entry point for running the command line interface.
+ */
+async function main() {
+  const worker = new Worker(APP_INSTANCE_PATH, {
+    workerData: argv.app_path,
+  });
+
+  worker.on('error', error => {
+    console.log(
```
Fleker, line 28: Would this be `console.error`?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export class CommandProcessor {
+  private worker: Worker;
+  constructor(worker: Worker) {
+    this.worker = worker;
+  }
+
+  // Wrapper for argument parsing with yargs.
+  private async parseIntent(userCommand: string): Promise<IntentMessage> {
+    const argv = await yargs()
```
Fleker, line 77: Not yet sure about all of these CLI parameter names

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+        }
+        return argv;
+      });
+
+    /**
+     * Validate arguments and return an IntentMessage
+     */
+    switch (argv.intent_type) {
+      case 'IDENTIFY':
+        if (argv.discovery_buffer === undefined) {
```
Fleker, line 131: Is it possible to make `identify` and `execute` separate commands, so that we can require options for each subcommand?

ie. `local-cli identify --discovery_buffer=...`

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+            'local_device_id is required to trigger an Execute intent'
+          );
+        }
+        if (argv.command === undefined) {
+          throw new Error('command is required to trigger an Execute intent');
+        }
+        if (argv.params === undefined) {
+          throw new Error('paramns is required to trigger an Execute intent');
+        }
+        if (argv.custom_data === undefined) {
```
Fleker, line 158: It's not necessarily required

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+      case 'EXECUTE':
+        if (argv.local_device_id === undefined) {
+          throw new Error(
+            'local_device_id is required to trigger an Execute intent'
+          );
+        }
+        if (argv.command === undefined) {
+          throw new Error('command is required to trigger an Execute intent');
+        }
+        if (argv.params === undefined) {
```
Fleker, line 155: Not every command has parameters.

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+        readlineInterface.question('Awaiting input...\n', async input => {
+          if (input.length === 0) {
+            resolve();
+            return;
+          }
+
+          //Parse the command.
+          const command = input.split(' ')[0];
+
+          if (input === 'exit') {
```
Fleker, line 201: May want to also include the `.exit` string used by the Node terminal environment.

src/index.ts
```diff
@@ -4,4 +4,7 @@ export * from './platform/stub-setup';
 export * from './platform/smart-home-app';
 export * from './platform/device-manager';
 export * from './platform/execute';
+/**
```
Fleker, line 7: It seems like a number of these changes are also in #39. Once that's added, it'll be good to rebase this, which may shrink the diff.

package.json
cdaly333: This update is currently based on #39, so it includes all of its changes.  Everything past 383d230 should exclude changes common to #39.  Will rebase once #39 is merged.

package.json
cdaly333: Catching all `.ts` files seems to catch the `.d.ts` files in `build/`.  I could do `eslint './test/**/*.ts' --ignore-pattern '**/*.d.ts'`, but I'm not sure that's much better than what I have now.  (This comment probably belongs in #39 but the diff is messy - my bad for making dependent PRs.  See description)

src/index.ts
cdaly333: Noted.  If you want to isolate just the CLI changes for now, I squashed all these changes so viewing 383d230 in GitHub will ignore the overlapping changes from #39 . (though this will change as I update this PR)

Fleker: Ack

src/cli/app-instance.ts
cdaly333: Changed to checking if false and throwing a new `Error`.

src/cli/main.ts
cdaly333: Done

src/cli/main.ts
cdaly333: It's not built in, but it should at least be a constant rather than a literal.  I added a declaration to `command-processor.ts`

src/cli/main.ts
cdaly333: Done

src/cli/command-processor.ts
cdaly333: I think splitting them into their separate commands will make these argument names less ambiguous.  Will do.

src/cli/command-processor.ts
cdaly333: I agree this would clean the parsing up.  Will do.

src/cli/command-processor.ts
cdaly333: Removed this check.

src/cli/command-processor.ts
cdaly333: Removed this requirement.

src/cli/command-processor.ts
```diff
@@ -0,0 +1,242 @@
...
+            // Set loop exit flag and return.
+            exit = true;
+            resolve();
+            return;
+          }
+
+          if (command !== 'send-intent') {
```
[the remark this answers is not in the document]
cdaly333: Will look into this.

src/cli/main.ts
cdaly333: Done.

src/cli/command-processor.ts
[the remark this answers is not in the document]
cdaly333: Done.

src/cli/command-processor.ts
cdaly333: I moved the `exit` command to yargs.

src/cli/command-processor.ts
cdaly333: Done.

src/cli/command-processor.ts
cdaly333: Split into commands.

package.json
```diff
@@ -8,22 +8,38 @@
   "license": "Apache-2.0",
   "scripts": {
     "compile": "tsc",
-    "lint": "tslint --project .",
     "test": "ava",
     "clean": "gts clean",
-    "fix": "gts fix"
+    "cli": "node ./build/src/cli/index.js",
```
proppy, line 13: you might consider having a separate `package.json`/project for the CLI.

src/cli/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Usage
+
+## To start the command line interface:
+
+Run `npm run cli -- --app_path APP_PATH`, where `APP_PATH` is an absolute path to the entry point of your app (typically `bundle.js` if basing on the [Local Home SDK Sample](https://github.com/actions-on-google/smart-home-local))
+
+## To send an IDENTIFY intent
+
+Run `identify IDENTIFY --request_id REQUEST_ID --discovery_buffer DISCOVERY_BUFFER --device_id DEVICE_ID`
```
proppy, line 13: did you consider having the tool take the identify/execute payload from stdin?

src/cli/app-instance.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * Worker thread that runs an instance of a mocked local fulfillment app.
+ */
+import {parentPort, workerData} from 'worker_threads';
```
proppy, line 4: curious why you decided to run this in a worker?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,225 @@
+import {Worker} from 'worker_threads';
```
proppy, line 1: the work interface seems generally useful to standardize the interface between the app and the platform w/ a message passing interface, did you consider adopting it for the library as well?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,225 @@
...
+ */
+export interface IntentMessage {
+  intentType: string;
+  requestId: string;
+}
+
+/**
+ * A class containing all parameters needed to process an Identify command.
+ */
+export class IdentifyMessage implements IntentMessage {
```
proppy, line 21: could we use the existing local home sdk types?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,225 @@
...
+    this.requestId = requestId;
+    this.discoveryBuffer = discoveryBuffer;
+    this.deviceId = deviceId;
+  }
+}
+
+/**
+ * A class containing all parameters needed to process an Identify command.
+ */
+export class ExecuteMessage implements IntentMessage {
```
proppy, line 43: ccould we use the existing local home sdk types?

test/platform/test-platform-execute.ts
```diff
@@ -33,33 +46,23 @@ async function registerDevice(
   Promise.resolve();
 }
 
-function createSimpleExecuteCommands(
```
proppy, line 36: is this related to this change (if yes, maybe update the description to list it)

src/cli/README.md
```diff
@@ -0,0 +1,36 @@
+# Command Line Interface
```
proppy, line 1: as a generate feedback, I think the CLI interface would be more useful if it was using a version of the platform stub that send real udp/tcp/http messages, that would allow developers to test interactively against their own device w/o deploying to a Google Home device.

src/cli/README.md
cdaly333: I considered this.  My thought process was if the developer is doing testing involving copying/pasting payload JSONS to test a handler, they may as well use the automated testing interface.  (i.e. I'm not sure copy/pasting payload JSONs is a "usable" CLI).  The idea for this PR was to provide a more intuitive / hotswappable command interface for sending intents on-demand and rapidly being able to tweak the arguments without needing to worry about schema compliance.  (I'm thinking of something like Postman's UI where you can simply tweak field values without needing to paste an entire JSON per request)

src/cli/app-instance.ts
cdaly333: I discussed with @Fleker doing the CLI synchronously vs using worker threads.  I think doing the intent fulfillment in a worker thread keeps things modular, which is good for adding features down the road.  For example, I could potentially swap out the command processor with a chrome UI.  I also figured it was good practice to keep the UI/View split up from the processing.

src/cli/README.md
proppy: I wonder if it would make sense to have a separate tool for generating those payload and piping them to CLI.

It seems to me that a general way to craft sync response/execute request from command line arguments cloud prove itself useful outside of the context of the Local Home SDK.

What do you think?

package.json
Fleker: I think we may want to set it up somehow to add the final index.js into the `bin` directory to run similar to `ava` and `gts`.

src/cli/README.md
Fleker: I think I'd agree to add in radios unless we use some sort of argument not to use radios (like `--simulate`) to suggest that the messages aren't real.

src/cli/README.md
```diff
@@ -0,0 +1,36 @@
...
+## To start the command line interface:
+
+Run `npm run cli -- --app_path APP_PATH`, where `APP_PATH` is an absolute path to the entry point of your app (typically `bundle.js` if basing on the [Local Home SDK Sample](https://github.com/actions-on-google/smart-home-local))
+
+## To send an IDENTIFY intent
+
+Run `identify IDENTIFY --request_id REQUEST_ID --discovery_buffer DISCOVERY_BUFFER --device_id DEVICE_ID`
+
+- `REQUEST_ID` is the request id.
+- `DISCOVERY_BUFFER` is the discovery buffer, formatted as a hex string.
```
Fleker, line 16: How do we come up with this hex string?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (error !== null) {
+          throw error;
+        }
+        return argv;
+      });
+
+    /**
+     * Validate arguments and return an IntentMessage
+     */
+    switch (argv._[0]) {
```
Fleker, line 141: Is `argv._` the right way to get the list of arguments?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,225 @@
...
+        return yargs
+          .option('request_id', {
+            describe: 'The request Id',
+            type: 'string',
+            demandOption: true,
+          })
+          .option('discovery_buffer', {
+            describe: 'The IDENTIFY dicovery buffer represented as a string',
+            type: 'string',
+            demandOption: false,
```
Fleker, line 95: If we assert below that the argument must be defined, shouldn't we `demandOption` here?

src/cli/README.md
cdaly333: I think one thing to consider is the actual use case / core issues that would solved by building such a tool.  If we find that developers are repeatedly creating JSON requests (if this is something that is regularly part of numerous different development workflows), then it may help to standardize this request generation.  I could see it being integrated as a simple web tool / integrated into the developer docs.

That being said, I think it's easy enough to integrate the request generation into tools that need it, on a case-by-case basis (what I'm doing right now).  So breaking it out as a separate tool may only complicate things.

As a side note, Postman integration could be interesting.  I could imagine[ importing a postman collection that we provide](https://learning.postman.com/docs/getting-started/importing-and-exporting-data/) as a way of providing a ready-to-use template of sample requests while piggybacking off a tool's UI/UX that's pretty universally loved.

src/cli/command-processor.ts
cdaly333: I'm not as compelled to use it for the test runner interface since everything is pretty synchronous and the stubs/platform exist within the app already.  For implementing actual radio functionality I could see it helping.

test/platform/test-platform-execute.ts
cdaly333: This was from #39 and was merged.  Should be fixed now.

package.json
cdaly333: Will look into this

src/cli/README.md
cdaly333: Will look into this

src/cli/README.md
cdaly333: That would depend on the device being identified (it is sent directly to the `identifyHandler`).  For the sample, for example, it would be a `cbor` encoded value.

src/cli/command-processor.ts
cdaly333: I believe so - there is usage in the yargs documentation.

src/cli/command-processor.ts
cdaly333: Done.  I removed the assertions, since they are redundant now that `demandOption` is per-command.

cdaly333:
Added UDP scan functionality. Adding EXECUTE radio support as well as tests/comments.  Still WIP status, will update this PR when it's ready for review.  I may open a new PR since the scope has changed (now adding radio functionality).

cdaly333:
Added UDP and TCP Execute write functionality.  Tested manually with sample device.  Still a work in progress.  Will open a new PR later today but wanted to push what I had for now.

cdaly333:
Squashed all changes since master and opened #47

src/cli/command-processor.ts
cdaly333: These are for formatting messages to pass to the worker, so they're different from the ones in the Local Home SDK.

src/cli/command-processor.ts
cdaly333: These are for formatting messages to pass to the worker, so they're different from the ones in the Local Home SDK.

src/cli/README.md
cdaly333: See #47

PR #48, opened 2020-08-05
Add command line interface with true UDP radio functionality

### LINKED ISSUE #37: Implement Node.js CLI for manual intent dispatch
Implement a basic Node CLI that allows sending intents via command line and observing output.

### LINKED ISSUE #46: Improve DeviceManagerStub implementation
See discussion from #42.  It's worth reconsidering the DeviceManagerStub, and exploring different ways of implementing radio responses (rather than just mocking).

cdaly333:
This PR adds true UDP radio functionality from Node for Identify and Execute requests as well as a CLI interface.

**Changes:**
- Added `CommandProcessor` class, which handles command line arguments and maintains a parent-worker loop to process commands until an exit is received.
- Added `udp-scan` command path, which broadcasts on `UDP` and forwards a response to the Identify handler fulfillment path.
- Added `simulate-identify` command path, which triggers the fulfillment's Identify handler with given parameters
- Added `trigger-execute` command path, which triggers an `ExecuteRequest`
- Added `PlatformWorker` class, which runs on a worker thread orchestrates CLI command processing across all stubs.
- Added stubs for `smarthome.DataFlow`
- Added `RadioController` class, which node handles radio communication protocols.
- Added `RadioDeviceManager`, which implements `smarthome.DeviceManager` with true radio fulfillment.
- Rename `DeviceManagerStub` to `MockDeviceManager`

**Notes:**
- I've manually tested functionality with the [ Local Home SDK Virtual Device](https://github.com/actions-on-google/smart-home-local/tree/master/device).  Some commands I used can be found in `/src/cli/README.md`

**Related Issues:**
- Progresses #2
- Resolves #46  
- Resolves #37

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,13 @@
     "compile": "tsc",
     "test": "ava",
     "clean": "gts clean",
+    "cli": "node ./build/src/cli/index.js",
     "lint": "eslint './src/**/*.ts' './test/**/*.ts'"
   },
   "dependencies": {
-    "ava": "^3.10.0"
+    "@types/yargs": "^15.0.5",
+    "ava": "^3.10.0",
+    "yargs": "^15.4.1"
   },
   "devDependencies": {
     "@ava/typescript": "^1.1.1",
@@ -29,6 +32,9 @@
     "gts": "^2.0.2",
     "prettier": "^2.0.5",
     "typescript": "^3.8.3"
+  },
+  "engines": {
+    "node": ">12.11.0"
   },
   "files": [
     "./build/src/**/*.ts",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "lint": "eslint './src/**/*.ts' './test/**/*.ts'"
   },
   "dependencies": {
-    "@types/yargs": "^15.0.5",
     "ava": "^3.10.0",
     "yargs": "^15.4.1"
   },
@@ -22,6 +21,7 @@
     "@ava/typescript": "^1.1.1",
     "@google/local-home-sdk": "^1.1.0",
     "@types/node": "^13.11.1",
+    "@types/yargs": "^15.0.5",
     "@typescript-eslint/eslint-plugin": "^3.7.0",
     "chalk": "^4.0.0",
     "eslint": "^7.5.0",
--- a/src/cli/command-processor.ts
+++ b/src/cli/command-processor.ts
@@ -187,7 +187,7 @@ export class CommandProcessor {
             const commandMessage = await this.parseIntent(input);
             if (commandMessage === undefined) {
               // Exit command recieved. Resolve promise.
-              console.log('Exit command recieved.  Terminating...');
+              console.log('Exit command received.  Terminating...');
               resolve();
               return;
             }
--- a/src/cli/command-processor.ts
+++ b/src/cli/command-processor.ts
@@ -1,0 +1,210 @@
+import {Worker} from 'worker_threads';
+import yargs from 'yargs/yargs';
+import * as readline from 'readline';
+import {CommandMessage, READY_FOR_MESSAGE, CHECK_READY} from './commands';
+import {ExecuteMessage, IdentifyMessage, ScanMessage} from './commands';
+import {UDPScanConfig} from '../radio/radio-controller';
+
+/**
+ * Class to parse raw user input from stdin.
+ */
+export class CommandProcessor {
+  private worker: Worker;
+  constructor(worker: Worker) {
+    this.worker = worker;
+  }
+
+  /**
+   * Parses a string and returns a CommandMessage, if valid.
+   * @param userCommand  A string to parse.
+   * @returns  A promise that resolves to a valid `CommandMessage`,
+   *   or void if an `exit` command was detected
+   */
+  private async parseIntent(
+    userCommand: string
+  ): Promise<CommandMessage | void> {
+    const argv = await yargs()
+      .command(
+        'udp-scan',
+        'Scan and identify devices with a given UDP scan configuration.',
+        yargs => {
+          return yargs
+            .option('broadcast_address', {
+              describe: 'Destination IP address foir the UDP broadcast.',
+              type: 'string',
+              demandOption: true,
+            })
+            .option('broadcast_port', {
+              describe: 'The destination port for the UDP discovery broadcast.',
+              type: 'number',
+              demandOption: true,
+            })
+            .option('listen_port', {
+              describe: 'The port to listen for the UDP discvery response.',
+              type: 'number',
+              demandOption: true,
+            })
+            .option('discovery_packet', {
+              describe: 'The payload to send in the UDP broadcast.',
+              type: 'string',
+              demandOption: true,
+            })
+            .option('request_id', {
+              describe: 'An optional request ID for the Identify Request.',
+              type: 'string',
+              default: 'default-request-id',
+              demandOption: false,
+            })
+            .option('device_id', {
+              describe: 'The device ID to include in the Identify Request.',
+              type: 'string',
+              default: 'default-device-id',
+              demandOption: false,
+            });
+        }
+      )
+      .command(
+        'simulate-identify',
+        'Manually trigger an Identify response.',
+        yargs => {
+          return yargs
+            .option('request_id', {
+              describe: 'The request Id',
+              type: 'string',
+              demandOption: true,
+            })
+            .option('discovery_buffer', {
+              describe: 'The IDENTIFY dicovery buffer represented as a string',
+              type: 'string',
+              demandOption: true,
+            })
+            .option('device_id', {
+              describe: 'The device Id',
+              type: 'string',
+              demandOption: true,
+            });
+        }
+      )
+      .command('trigger-execute', 'Trigger an Execute command.', yargs => {
+        return yargs
+          .option('local_device_id', {
+            describe: 'The local device Id',
+            type: 'string',
+            demandOption: true,
+          })
+          .option('command', {
+            describe: 'The execute command to send to device_id',
+            type: 'string',
+            demandOption: true,
+          })
+          .option('params', {
+            describe:
+              'The params argument for the execute command, in JSON format',
+            type: 'string',
+            default: '{}',
+            demandOption: false,
+          })
+          .option('custom_data', {
+            describe:
+              'The customData argument for the execute command, in JSON format',
+            type: 'string',
+            default: '{}',
+            demandOption: false,
+          });
+      })
+      .command('exit', 'Exit the command line interface.')
+      .demandCommand(1, 'Must provide a valid command')
+      .help()
+      .parse(userCommand, (error: Error) => {
+        if (error !== null) {
+          throw error;
+        }
+      });
+
+    /**
+     * Parse argv based on command and return a `CommandMessage`
+     */
+    const command = argv._[0];
+    switch (command) {
+      case 'udp-scan': {
+        const scanConfig = new UDPScanConfig(
+          argv.broadcast_address,
+          argv.broadcast_port,
+          argv.listen_port,
+          argv.discovery_packet
+        );
+        return new ScanMessage(argv.device_id, argv.request_id, scanConfig);
+      }
+      case 'simulate-identify':
+        return new IdentifyMessage(
+          argv.request_id,
+          argv.discovery_buffer,
+          argv.device_id
+        );
+      case 'trigger-execute':
+        return new ExecuteMessage(
+          argv.request_id,
+          argv.local_device_id,
+          argv.command,
+          JSON.parse(argv.params),
+          JSON.parse(argv.custom_data)
+        );
+      case 'exit':
+        return Promise.resolve();
+      default:
+        throw new Error('Unsupported command: ' + command);
+    }
+  }
+
+  /**
+   * The main user input to response loop.
+   * Recieves raw user input and attempts to parse and forward
+   * messages to the worker thread.
+   */
+  async processUserInput(): Promise<void> {
+    // Open a readline interface
+    const readlineInterface = readline.createInterface({
+      input: process.stdin,
+      output: process.stdout,
+    });
+
+    await new Promise(resolve => {
+      // Pings the worker thread.
+      this.worker.postMessage(CHECK_READY);
+
+      this.worker.on('message', async message => {
+        // Waits for worker thread to indicate it's ready to take another message.
+        if (message !== READY_FOR_MESSAGE) {
+          return;
+        }
+        readlineInterface.question('Awaiting input...\n', async input => {
+          if (input.length === 0) {
+            // Ping worker thread.
+            this.worker.postMessage(CHECK_READY);
+            return;
+          }
+          try {
+            const commandMessage = await this.parseIntent(input);
+            if (commandMessage === undefined) {
+              // Exit command recieved. Resolve promise.
+              console.log('Exit command recieved.  Terminating...');
+              resolve();
+              return;
+            }
+            // Post `CommandMessage` to worker thread.
+            this.worker.postMessage(commandMessage);
+          } catch (error) {
+            // Catch possible parsing error.
+            console.error(error);
+            // Ping worker thread.
+            this.worker.postMessage(CHECK_READY);
+          }
+        });
+      });
+    });
+
+    // Clean up the readline interface.
+    readlineInterface.close();
+    return Promise.resolve();
+  }
+}
--- a/src/cli/command-processor.ts
+++ b/src/cli/command-processor.ts
@@ -30,7 +30,7 @@ export class CommandProcessor {
         yargs => {
           return yargs
             .option('broadcast_address', {
-              describe: 'Destination IP address foir the UDP broadcast.',
+              describe: 'Destination IP address for the UDP broadcast.',
               type: 'string',
               demandOption: true,
             })
@@ -74,7 +74,7 @@ export class CommandProcessor {
               demandOption: true,
             })
             .option('discovery_buffer', {
-              describe: 'The IDENTIFY dicovery buffer represented as a string',
+              describe: 'The IDENTIFY discovery buffer represented as a string',
               type: 'string',
               demandOption: true,
             })
--- a/src/radio/dataflow.ts
+++ b/src/radio/dataflow.ts
@@ -1,0 +1,53 @@
+/**
+ * Stubs to allow handlers to access smarthome.DataFlow
+ */
+export const DataFlowStub = {
+  UdpRequestData: class {
+    protocol: smarthome.Constants.Protocol = smarthome.Constants.Protocol.UDP;
+    requestId = '';
+    deviceId = '';
+    data = '';
+    port = 0;
+  },
+};
+
+/**
+ * An implementation of `smarthome.DataFlow.UdpResponseData` for
+ * responding to UDP requests.
+ */
+export class UdpResponseData implements smarthome.DataFlow.UdpResponseData {
+  udpResponse: smarthome.DataFlow.UdpResponse;
+  requestId: string;
+  deviceId: string;
+  protocol: smarthome.Constants.Protocol = smarthome.Constants.Protocol.UDP;
+  /**
+   * @param requestId  The requestId of associated `UdpRequestData`.
+   * @param deviceId  The id of the device that the request was sent to.
+   * @param udpResponse  The contents of the response.
+   * @returns  A new UdpResponseData instance
+   */
+  constructor(
+    requestId: string,
+    deviceId: string,
+    udpResponse: smarthome.DataFlow.UdpResponse
+  ) {
+    this.requestId = requestId;
+    this.deviceId = deviceId;
+    this.udpResponse = udpResponse;
+  }
+}
+
+/**
+ * An implementation of `smarthome.DataFlow.UdpResponse` for
+ * responding to UDP requests.
+ */
+export class UdpResponse implements smarthome.DataFlow.UdpResponse {
+  responsePackets?: string[];
+  /**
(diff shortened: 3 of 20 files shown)


src/radio/dataflow.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Stubs to allow handlers to access smarthome.DataFlow
+ */
+export const DataFlowStub = {
```
Fleker, line 4: Will this just be for UDP or will the same file / `DataFlowStub` include other radios?

src/radio/radio-controller.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @param broadcastAddress  The destination UDP broadcast address.
+   * @param broadcastPort  The destination UDP broadcast port.
+   * @param listenport  The listen port for the UDP response.
+   * @param discoveryPacket  The payload to send in the UDP broadcast.
+   * @returns  A new `UDPScanConfig` instance.
+   */
+  constructor(
+    broadcastAddress: string,
+    broadcastPort: number,
+    listenport: number,
```
Fleker, line 29: We could make this camelCase for consistency with `broadcastPort`.

src/radio/radio-controller.ts
```diff
@@ -0,0 +1,166 @@
...
+  address: string;
+}
+
+/**
+ * A class to contain all Node radio functionality.
+ */
+export class RadioController {
+  /**
+   * A helper function to create timeout promises.
+   * @returns  A new Promise that resolves after RADIO_TIMEOUS milliseconds.
```
Fleker, line 53: RADIO_TIMEOUT

src/radio/radio-controller.ts
```diff
@@ -0,0 +1,166 @@
+import * as dgram from 'dgram';
+import {UdpResponse} from './dataflow';
+
+/**
+ * A default radio timeout, in milliseconds.
+ */
+const RADIO_TIMEOUT = 2500;
```
Fleker, line 7: Is there a way to change this default? Is `2500ms` the right timeout?

src/radio/radio-controller.ts
```diff
@@ -0,0 +1,166 @@
...
+      // Decode the discovery packet and send it.
+      const payload = Buffer.from(udpScanConfig.discoveryPacket, 'hex');
+      socket.send(
+        payload,
+        udpScanConfig.broadcastPort,
+        udpScanConfig.broadcastAddress,
+        error => {
+          if (error !== null) {
+            throw new Error(
+              'Failed to send UDP discovery packet:' + error.message
```
Fleker, line 92: I think you can do string templating here, ie. \``Failed to send UDP discovery packet: ${error.message}\``

Fleker: But with one backtick on each side.

src/radio/radio-controller.ts
```diff
@@ -0,0 +1,166 @@
...
+ * A class to contain all Node radio functionality.
+ */
+export class RadioController {
+  /**
+   * A helper function to create timeout promises.
+   * @returns  A new Promise that resolves after RADIO_TIMEOUS milliseconds.
+   */
+  private createTimeoutPromise(): Promise<void> {
+    return new Promise(resolve => {
+      setTimeout(resolve, RADIO_TIMEOUT);
```
Fleker, line 57: Would it make more sense for this to reject rather than resolve?

src/radio/radio-device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+   * @returns  A promise that resolves to the determined UdpResponseData.
+   */
+  private async processUdpRequestData(
+    udpRequestData: smarthome.DataFlow.UdpRequestData
+  ): Promise<smarthome.DataFlow.UdpResponseData> {
+    const payload = Buffer.from(udpRequestData.data, 'hex');
+    const localAddress = this.deviceIdToAddress.get(udpRequestData.deviceId);
+    const udpResponse = await this.radioController.sendUdpMessage(
+      payload,
+      localAddress!,
```
Fleker, line 61: Do we need to assert it's not undefined? What happens if it is undefined

src/radio/dataflow.ts
cdaly333: This will include other radio stubs as well, one for each exported class in `smarthome.DataFlow` - `UdpRequestData`, `TcpRequestData`, and `HttpRequestData`

src/radio/radio-controller.ts
cdaly333: I think it makes sense for the timeout promise to resolve, since that's its purpose.  The `Promise.race()` call rejects after this timeout resolves, which I think is the appropriate place to start the `reject` since up until then it's expected/normal behavior

src/radio/radio-device-manager.ts
cdaly333: It should be impossible for it to be undefined, since we check that the address is in the map in the `send` call.

(May be unrelated, since I'm not sure if the TypeScript compiler would be able to relate these calls across functions anyways, but there's an open issue where `map.has()` doesn't guarantee `map.get()` is not `undefined`: https://github.com/microsoft/TypeScript/issues/9619)

src/radio/radio-controller.ts
cdaly333: Done.

src/radio/radio-controller.ts
cdaly333: Done

src/radio/radio-controller.ts
cdaly333: It's an arbitrary timeout value - I'm not sure what would be the optimal value would be for default but I'm assuming 2500ms would be safe in most cases.  I added the ability to specify timeout.

src/radio/radio-controller.ts
cdaly333: Done

cdaly333: Put quotes instead of backticks.  Will fix.

cdaly333: Done

cdaly333:
Fixed some changes mentioned in #49

src/radio/radio-controller.ts
```diff
@@ -50,11 +50,11 @@ export interface UDPScanResults {
 export class RadioController {
   /**
    * A helper function to create timeout promises.
-   * @returns  A new Promise that resolves after RADIO_TIMEOUS milliseconds.
+   * @returns  A new Promise that resolves after RADIO_TIMEOUT milliseconds.
    */
-  private createTimeoutPromise(): Promise<void> {
+  private createTimeoutPromise(timeout: number): Promise<void> {
```
Fleker, line 55: Should add a `@param timeout`.

src/radio/radio-controller.ts
```diff
@@ -63,15 +63,18 @@ export class RadioController {
    * @param udpScanConfig  A scan configuration containing UDP parameters.
    * @return  A promise that resolves to the determined `UDPScanResults`
    */
-  public async udpScan(udpScanConfig: UDPScanConfig): Promise<UDPScanResults> {
+  public async udpScan(
+    udpScanConfig: UDPScanConfig,
+    timeout = RADIO_TIMEOUT
```
Fleker, line 68: Should add an `@param timeout`.

src/radio/radio-controller.ts
```diff
@@ -99,12 +102,10 @@ export class RadioController {
...
     return Promise.race([
       discoveryBuffer,
-      this.createTimeoutPromise().then(() => {
+      this.createTimeoutPromise(timeout).then(() => {
         // Close the socket if timed out
         socket.close();
-        throw new Error(
-          'UDP scan timed out after ' + RADIO_TIMEOUT.toString() + 'ms.'
-        );
+        throw new Error(`UDP scan timed out after ${RADIO_TIMEOUT}ms.`);
```
Fleker, line 108: If you're using `timeout` as a param then we should put that in the error message.

src/radio/radio-controller.ts
```diff
@@ -122,7 +123,8 @@ export class RadioController {
     address: string,
     port: number,
     listenPort: number,
-    expectedResponsePackets = 0
+    expectedResponsePackets = 0,
+    timeout = RADIO_TIMEOUT
```
Fleker, line 127: Should add a `@param timeout`.

src/radio/radio-controller.ts
```diff
@@ -155,11 +157,9 @@ export class RadioController {
     // Timeout if still waiting for a response.
     return Promise.race([
       discoveryBuffer,
-      this.createTimeoutPromise().then(() => {
+      this.createTimeoutPromise(timeout).then(() => {
         socket.close();
-        throw new Error(
-          'UDP send timed out after ' + RADIO_TIMEOUT.toString() + 'ms.'
-        );
+        throw new Error(`UDP send timed out after ${RADIO_TIMEOUT} ms.`);
```
Fleker, line 162: Should use `${timeout}ms`.

test/radio/radio-controller-test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Class to manage a local UDP server
+ */
+class UdpServer {
+  private server: dgram.Socket | undefined;
+  public startServer(listenPort = 3112) {
+    this.server = dgram.createSocket('udp4');
+    this.server.on('message', (message: Buffer, rinfo) => {
+      this.server!.send(UDP_DUMMY_BUFFER_1, rinfo.port, rinfo.address);
```
Fleker, line 17: Do we need to assert that the server is not undefined? What would happen if it is undefined?

package.json
```diff
@@ -10,10 +10,13 @@
     "compile": "tsc",
     "test": "ava",
     "clean": "gts clean",
+    "cli": "node ./build/src/cli/index.js",
     "lint": "eslint './src/**/*.ts' './test/**/*.ts'"
   },
   "dependencies": {
-    "ava": "^3.10.0"
+    "@types/yargs": "^15.0.5",
```
Fleker, line 17: I don't think you need `@types/yargs` as a dependency as the types are only used for development. It can go in `devDependencies`.

test/radio/radio-controller-test.ts
```diff
@@ -0,0 +1,81 @@
+import test from 'ava';
+import * as dgram from 'dgram';
```
Fleker, line 2: Is `dgram` a dependency that needs to be included in `package.json`?

cdaly333: `dgram` is a core Node module so there's no need to include it as a dependency.

test/radio/radio-controller-test.ts
cdaly333: As far as I can tell it is impossible that `this.server` will be undefined at this point, since it's a private member that is definitely set in the first line of `startServer`, and never set back to `undefined`.

package.json
cdaly333: Done

src/radio/radio-controller.ts
cdaly333: Done.

src/radio/radio-controller.ts
cdaly333: Done

src/radio/radio-controller.ts
cdaly333: Done

src/radio/radio-controller.ts
cdaly333: Done

src/radio/radio-controller.ts
cdaly333: Done

src/radio/radio-device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * The default port to listen on for radio responses.
+ */
+const DEFAULT_LISTEN_PORT = 3311;
+
+/**
+ * An implementation of `smarthome.DeviceManager` that implements radio functionality.
+ */
+export class RadioDeviceManager implements smarthome.DeviceManager {
```
proppy, line 12: Curious if you think it would make sense to move the mocking abstraction one-level below at the controller level and have a common DeviceManager stub for both?

Alternative did you consider folding the controller implement as private method of the manager, or if you want to keep the level of indirection split them away in per-protocol implementation?

cdaly333: I think it's worth keeping them separate. I can see two separate use cases:

1. Automated testing using the stubs to place their fulfillment app as the unit under test.  I think in this scenario, we want to restrict the testing context to the handlers themselves as much as possible.  Mocking the radio instead of the `DeviceManager.send()` call would force the developer to think about how the Local Home Platform interprets transport protocol response commands for _every_ unit test involving commands.

2. Manual testing across radio with command line.  In this scenario, the developer doesn't want to mock at all, so merging would not benefit here.  Additionally, **this** is the scenario where it would make sense for the developer to consider how the Local Home Platform interprets their radio responses, since they are doing more integrated testing.

This is just my two cents.

Re: folding `RadioController` into `RadioDeviceManager`, I think it's worth keeping them separate, and with all protocols in one `RadioController`.  It might be worth splitting the protocol implementations into individual classes that are managed within `RadioController`.  In any case I don't think the `RadioDeviceManager` stub is the right place to manage the individual protocol implementations.  I think the `RadioDeviceManager` should just be a middleman that handles translation of handler requests to radio requests, and radio responses to handler responses, with access to a radio API that it can assume works.

src/radio/radio-device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+  private radioController: RadioController;
+  private listenPort: number;
+  public deviceIdToAddress: Map<string, string> = new Map<string, string>();
```
proppy, line 15: does this need to be public?

src/radio/radio-device-manager.ts
proppy: >  Mocking the radio instead of the DeviceManager.send() call would force the developer to think about how the Local Home Platform interprets transport protocol response commands for every unit test involving commands.

One thing to consider is that with the current implementation we pollute the public interface of `DeviceManager` to enable the unit testing usecase, while we strictly implement the `DeviceManager` for the integration testing usecase.

Currently it seems that the `RadioDeviceManager` does the following:
a. resolve the address from deviceId
b. pass the response port
c. forward UdpRequestData as exploded arguments
d. return UdpResponseData

I'm curious if you think those still be useful in a unit testing context?
- a. could be useful to validate than the id match one that was actually registered
- b. could be a placeholder
- c. and d. could simply forward the command/response to a recording stub.

What we force developer to do really depends on the API that we surface back to the unittest, and in that case it seems that what they would need is:
- checking which *RequestData was passed thru
- specifying which *ResponseData to return

Which could fall within the responsability of the `RadioController` if we adapted its calling convention to consume platform message rather than exploded arguments.

I also wonder if we could simply let the developer specific `sinon.js` [fake](https://sinonjs.org/releases/latest/fakes/) and reuse their expectation interface.

>  I think the RadioDeviceManager should just be a middleman that handles translation of handler requests to radio requests, and radio responses to handler responses, with access to a radio API that it can assume works

This is very sound to isolate the translation layer, but currently it does seems that we are translating twice by introducing an additional interface:
```
DeviceManager *RequestData + resolve device endpoint -> RadioController.sendUdpMessage arguments -> node.js dgram API call
```
Where as we really only need to translate:
```
DeviceManager *RequestData + resolved device endpoint -> node.js dgram API call
```
>  It might be worth splitting the protocol implementations into individual classes that are managed within RadioController

Do you think those could be manager by the DeviceManager? since it will also contain the logic to downcast the command to the right RequestData type.

cdaly333: > One thing to consider is that with the current implementation we pollute the public interface of DeviceManager to enable the unit testing usecase, while we strictly implement the DeviceManager for the integration testing usecase.

My perspective is we don't want the developer to speculate about the internals of `RadioController`/`DeviceManager`, since we want to present the class as a blackbox that fulfills radio commands, such that they can test how their handlers react.

I don't know that it would be helpful for the developer to need to mock `*RequestData`, `*ResponseData`.  They would be mocking structs that are used a couple layers down the call stack of `DeviceManager.send()`, when ideally they can ignore those layers and just mock the highest-level call directly.

That being said, I see how requiring the developer to work around the `MockDeviceManager` interface is a similar issue.  I think `sinon.js` might be worth looking into, so I opened issue #53.  It would allow the developer to pick and choose precisely what they want to mock, instead of imposing our interface.

>This is very sound to isolate the translation layer, but currently it does seems that we are translating twice by introducing an additional interface:

My perspective is we do one major translation: LHSDK-domain arguments to a protocol-domain arguments.  The `RadioController` just forwards the translated arguments to a fulfillment class, which plugs them into the respective APIs. 

A couple other things:
- For the chrome version, I'll be able to swap out the `RadioController` for a chrome-specific radio implementation, without re-translating
- After implementing UDP, and TCP, the `RadioDeviceManager` class could be well over 500 lines if implementing everything together, which starts to eat into readability. 

>Do you think those could be manager by the DeviceManager? since it will also contain the logic to downcast the command to the right RequestData type.

This is another option that I will explore if classes become too bloated.  I don't know that we need the flexibility of swapping out protocol-specific implementation classes.

src/cli/command-processor.ts
```diff
@@ -0,0 +1,210 @@
...
+    userCommand: string
+  ): Promise<CommandMessage | void> {
+    const argv = await yargs()
+      .command(
+        'udp-scan',
+        'Scan and identify devices with a given UDP scan configuration.',
+        yargs => {
+          return yargs
+            .option('broadcast_address', {
+              describe: 'Destination IP address foir the UDP broadcast.',
```
ToniCorinne, line 33: "for"

src/cli/command-processor.ts
```diff
@@ -0,0 +1,210 @@
...
+  private async parseIntent(
+    userCommand: string
+  ): Promise<CommandMessage | void> {
+    const argv = await yargs()
+      .command(
+        'udp-scan',
+        'Scan and identify devices with a given UDP scan configuration.',
+        yargs => {
+          return yargs
+            .option('broadcast_address', {
```
Fleker, line 32: If the commands are using kebab-case, should the options change to use that style to be consistent?

src/cli/command-processor.ts
```diff
@@ -0,0 +1,210 @@
...
+        'Manually trigger an Identify response.',
+        yargs => {
+          return yargs
+            .option('request_id', {
+              describe: 'The request Id',
+              type: 'string',
+              demandOption: true,
+            })
+            .option('discovery_buffer', {
+              describe: 'The IDENTIFY dicovery buffer represented as a string',
```
Fleker, line 77: discovery

src/radio/radio-device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * @param  radioController  The radio controller used for radio communication.
+   * @param  listenPort  The port to listen on for radio responses.
+   */
+  constructor(
+    radioController: RadioController,
+    listenPort: number = DEFAULT_LISTEN_PORT
```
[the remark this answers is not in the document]
cdaly333: This is only for UDP read.  We access the `RadioController` directly from `PlatformWorker` to set the `listenPort` for a UDP broadcast.

src/cli/command-processor.ts
cdaly333: I noticed the Local Home Sample device options used [underscore naming](https://github.com/actions-on-google/smart-home-local/tree/master/device), and `package.json` script commands seem to be named with kebab-case.  I think we should keep the options with underscores to follow that pattern.  If anything, we could change the command name to use underscores as well, but I personally think the current naming works well.  If you feel strongly I will change.

src/radio/radio-device-manager.ts
cdaly333: Made private.

src/cli/command-processor.ts
cdaly333: Done

src/cli/command-processor.ts
cdaly333: Done.  Will see if there is some global spellcheck I can run for comments.

src/radio/radio-device-manager.ts
proppy: > I don't know that it would be helpful for the developer to need to mock *RequestData, *ResponseData

Sorry for the confusion, I wasn't suggestion that developer would need to mock those, but rather that we (or they) could provide a mock `RadioController` that capture them, rather than surfacing dedicated assertions method on the an implementation of the `DeviceManager` platform interface (we could discuss this in #53 if you think that's orthogonal to this change).

>  The RadioController just forwards the translated arguments to a fulfillment class

My understanding was that in its current implementation the `RadioController` was the fulfilment, did you mean `RadioDeviceManager` ?

> I don't know that we need the flexibility of swapping out protocol-specific implementation classes.

My instinct was that swapping them would be useful for testing (setup fake radio, verify it sent and received the right packets) but that could be discussed as part of #53

cdaly333: >Sorry for the confusion, I wasn't suggestion that developer would need to mock those, but rather that we (or they) could provide a mock RadioController that capture them, rather than surfacing dedicated assertions method on the an implementation of the DeviceManager platform interface (we could discuss this in #53 if you think that's orthogonal to this change).

I think I understand what you're saying (I'm probably also being a bit unclear with terminology).  My main question was what is the advantage of a `MockRadioController` (or individual radio mock classes.
) vs a `MockDeviceManager`.

>My understanding was that in its current implementation the RadioController was the fulfilment, did you mean RadioDeviceManager ?

Yes, updated my previous comment.

I think the main takeaway is to not reinvent the wheel and consider #53 for mocking.  We can continue discussion in that issue if there's anything else.

PR #55, opened 2020-08-17
Split single package into separate `cli/` and `stubs/` packages

cdaly333:
This PR splits up command line UI layer and the core stubs/radio implementation into their own separate packages.  This achieves a couple main things:

- **Abstracts stubs from UI to enable the modularize UI layer.  This makes way for the chrome implementation.**
- Removed `yargs` dependency from stub package. 

The CLI package imports the stub package and implements a UI layer on top of it.

**Core changes:**
- Added `stubs/` and `cli` directories with new node packages.
- Added `RadioController` interface. (to allow abstract radio controller implementations - namely the chrome version)
- Renamed `RadioController` class to `NodeRadioController`.
- Remove `yargs` dependency from core stub package.
- Moved `UDPScanConfig`, `RemoteInfo` and `UDPScanResults` to `dataflow.ts`.
- Removed redundant dependencies from `package.json`s 

**Minor tweaks:**
- Minor `README.md` updates.
- Remove some unneeded `console.log` calls.

**Issues:**
- Progresses #11


cli/src/command-processor.ts
```diff
@@ -3,7 +3,7 @@ import yargs from 'yargs/yargs';
 import * as readline from 'readline';
 import {CommandMessage, READY_FOR_MESSAGE, CHECK_READY} from './commands';
 import {ExecuteMessage, IdentifyMessage, ScanMessage} from './commands';
-import {UDPScanConfig} from '../radio/radio-controller';
+import {UDPScanConfig} from 'local-home-testing/build/src/radio';
```
Fleker, line 6: Do they need to come from `/build` or can the values be pulled directly from the Typescript source?

stubs/src/radio/radio-controller.ts
```diff
@@ -1,63 +1,58 @@
...
 /**
  * A default radio timeout, in milliseconds.
  */
 const RADIO_TIMEOUT = 2500;
-
 /**
- * A class to contain all parameters required to perform
- * a UDP scan.
+ * A helper function to create timeout promises.
+ * @returns  A new Promise that resolves after RADIO_TIMEOUT milliseconds.
```
Fleker, line 10: This doesn't resolve after `RADIO_TIMEOUT` but after a parameter.

cli/README.md
```diff
@@ -8,11 +8,11 @@ TODO(cjdaly) update usage
```
proppy, line 8: maybe do those in a separate PR?

cli/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "local-home-testing-cli",
+  "version": "0.0.1",
+  "description": "A command-line interface for interacting with Local Home SDK stubs.",
```
proppy, line 4: is that actually using the stubs of the new radio interfaces?

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "private": true,
```
proppy, line 5: should that still be private?

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "author": "Google LLC.",
+  "license": "Apache-2.0",
+  "main": "./build/src/index.js",
+  "scripts": {
+    "compile": "tsc",
+    "test": "ava",
+    "clean": "gts clean",
+    "cli": "node ./build/src/index.js",
```
proppy, line 13: ts-node?

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "lint": "eslint './src/**/*.ts' './test/**/*.ts'"
```
proppy, line 14: should that be `gts check` ? (i.e: what result of `gts init`)

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "devDependencies": {
+    "ava": "^3.10.0",
+    "@ava/typescript": "^1.1.1",
+    "@google/local-home-sdk": "^1.1.0",
+    "@types/node": "^13.11.1",
+    "@types/yargs": "^15.0.5",
+    "@typescript-eslint/eslint-plugin": "^3.7.0",
+    "chalk": "^4.0.0",
+    "eslint": "^7.5.0",
```
proppy, line 28: shouldn't that come for free with gts?

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "esm": "^3.2.25",
+    "gts": "^2.0.2",
+    "prettier": "^2.0.5",
+    "typescript": "^3.8.3"
+  },
+  "engines": {
+    "node": ">12.11.0"
+  },
+  "files": [
+    "./build/src/**/*.ts",
```
proppy, line 41: should this point to the *.js files?

cli/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "./build/test/example/**/*.ts"
+  ],
+  "ava": {
+    "files": [
+      "test/**/*",
+      "!test/**/*fixtures.ts"
+    ],
+    "require": [
+      "esm"
```
proppy, line 50: is that required, I don't see it in https://github.com/avajs/typescript

cli/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "local-home-testing-cli",
```
proppy, line 2: @google namespace?

cli/src/command-processor.ts
```diff
@@ -87,6 +87,12 @@ export class CommandProcessor {
       )
       .command('trigger-execute', 'Trigger an Execute command.', yargs => {
         return yargs
+          .option('request_id', {
```
proppy, line 90: should that go in a separate change?

cli/src/platform-worker.ts
```diff
@@ -59,7 +64,7 @@ export class PlatformWorker {
           // Trigger the identifyHandler with scan results.
           await this.mockLocalHomePlatform.triggerIdentify(
             scanMessage.requestId,
-            scanRemoteInfo.buffer,
+            Buffer.from(scanRemoteInfo.scanData, 'hex'),
```
proppy, line 67: what about doing `hex`/`Buffer` related bug fixes in a separate change?

stubs/src/radio/radio-controller.ts
```diff
@@ -142,24 +137,26 @@ export class RadioController {
...
-        socket.send(payload, port, address, error => {
-          if (error) {
-            throw new Error(`Failed to send UDP message: ${error.message}`);
-          }
-          console.log('Sent UDP message: ', payload);
-          if (expectedResponsePackets === 0) {
-            // Resolve early if we aren't expecting any response.
-            resolve(new UdpResponse());
-          }
+        socket.bind(listenPort, () => {
```
proppy, line 140: curious what this fix is about?

cli/src/command-processor.ts
cdaly333: Since the TypeScript compiler doesn't resolve the `local-home-testing` local dependency, I believe this import gets resolved at runtime, so I can't target the typescript source.  One option is using `tsconfig` path mapping to hard code the TypeScript source but I'm unsure that would be much better so will leave as-is for now.

stubs/src/radio/radio-controller.ts
cdaly333: Fixed.

cli/README.md
cdaly333: Will update `README`s in a separate PR.

cli/package.json
cdaly333: The `local-home-testing-cli` project imports the stubs and true radio functionality from `local-home-testing`, yes.

cli/package.json
cdaly333: Removed private.

cli/package.json
cdaly333: Since we're not using `ts-node` in other projects, I'm inclined to just use `node`.  If I'm not mistaken, the only difference would be the compilation step (and we `include` the build files).

cli/package.json
cdaly333: Went through both projects and removed most the `eslint` dependencies.

As a side note, some `eslint` dependencies seem to be a bit outdated in `gts`.  In particular, I had to include the latest version of `@typescript-eslint/eslint-plugin` since it was seemingly required to recognize the Local Home SDK types through triple-slash directives.

cli/package.json
cdaly333: Removed.

cli/package.json
cdaly333: Fixed

cli/package.json
cdaly333: Not required. Removed.

cli/package.json
cdaly333: Done.

cli/src/command-processor.ts
cdaly333: Will include in separate PR.

cli/src/platform-worker.ts
cdaly333: Will do.

stubs/src/radio/radio-controller.ts
cdaly333: It's to ensure `socket.send()` is called after `socket.bind()`.  If I'm not mistaken, `socket.send()` will still work without the socket being bound, but it will resort to a default unwanted behavior.  I'll see if I can move this into another PR.

cdaly333:
Removed unrelated changes.  Leaving a few comments unresolved until I open new PRs for the unrelated changed.  All other comments should be addressed.

cli/src/command-processor.ts
cdaly333: Done

stubs/src/radio/radio-controller.ts
cdaly333: Done. See #60

cli/src/platform-worker.ts
cdaly333: Done. see #59

cli/package.json
proppy: This does make the `cli` script dependent on `compile`, should we automatically run it before? or is it documented in the README? (one thing I'm afraid is that since the `cli` script is mainly used for development, one my forget to recompile between each modification).

ToniCorinne:
@cdaly333 got some merge conflicts on here and on #56  and #59 as well - let me know the order in which you want to handle getting those merged.

cdaly333:
Resolved merge conflicts here.  We should merge this one first, followed by #59, and finally #56.  Will need to `rebase` after each merge from here on.

PR #56, opened 2020-08-17
Add Chrome implementation with GUI and proxied radio fulfillment

### LINKED ISSUE #2: Implement virtual and real device integration

### LINKED ISSUE #9: Implement virtual device integration - Chrome

### LINKED ISSUE #11: Implement interactive stub library in Chrome

cdaly333:
**Note: this PR depends on PR #55.  Too isolate the changes after #55, see all commits after and including commit 34be465** 

This PR introduces a new package that implements the Local Home SDK stubs in a Chrome environment.

- The stubs are bundled using `webpack` for compatibility.
- A controller layer provides hooks to coordinate platform actions.
- A proxy server runs on chrome and listens on a `WebSocket` for radio commands.
- A proxy `RadioController` implementation fulfills radio commands through the proxy server.
- A UI layer allows I/O through a DOM interface.

**General flow:**
1. DOM GUI Input
2. `ProxyRadioClient` creates `ProxyRequest` and sends over `WebSocket`
3. `ProxyRadioServer` parses `ProxyRequest` and fulfills with `NodeRadioController`
4. `ProxyRadioServer` creates `ProxyResponse`and sends over `WebSocket`
5. `ProxyRadioClient` handles `ProxyResponse`
6. DOM GUI Output
 
**Specific changes (ordered by significance):**
- Created a new `npm` package in `chrome/` to encompass Chrome app.
- Added `ProxyRadioClient` which implements `RadioController` by fulfilling radio requests over a `WebSocket` connection.
- Added `ProxyRadioServer` which uses a `NodeRadioController` to fulfill radio requests for a `ProxyRadioClient` over a `WebSocket` connection.
- Added `PlatformCoordinator` class which manages the `MockLocalHomePlatform` and radio functionality.
- Added a simple HTML/CSS GUI.
- Add `webpack` configuration to package the app for chrome.
- Added `radio-message.ts` which types proxy messages.
- Added a basic `README.md` for the chrome app.

**Issues:**
- Resolves #9 
- Resolves #2 
- Resolves #11

**Notes:**
- Has been manually tests but needs automatic tests.

**Visual of the Chrome GUI:**
image


chrome/src/browser/proxy-client.ts
```diff
@@ -0,0 +1,130 @@
...
+export class ProxyRadioClient implements RadioController {
+  private webSocket: WebSocket;
+  private onUdpScanResponse: ProxyResponseAction<UdpScanResponse> | undefined;
+  private onUdpSendResponse: ProxyResponseAction<UdpSendResponse> | undefined;
+
+  /**
+   * Establishes the local WebSocket connection.
+   * @returns  A new `ProxyRadioClient` instance.
+   */
+  constructor(port = 4321) {
```
Fleker, line 30: `@param port`?

chrome/src/browser/proxy-client.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Sends a `UdpScanRequest` over the WebSocket connection
+   * and resolves with the response.
+   * @param payload  The UDP message payload.
+   * @param address  The UDP destination address.
+   * @param port  The UDP destination port.
+   * @param listenPort  The UDP port to listen on.
+   * @param expectedResponsePackets  The amount of response packets to listen for.
+   * @param timeout
```
Fleker, line 101: `timeout` doesn't seem to be a parameter here

chrome/src/node/proxy-server.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * A WebSocket server that runs on node to provide radio fulfillment.
+ * Uses the `NodeRadioController` implementation to fulfill requests.
+ */
+export class ProxyRadioServer {
+  private webSocketServer: ws.Server;
+  private nodeRadioController: NodeRadioController;
+
+  constructor(port = 4321) {
```
proppy, line 19: should that be configurable?

chrome/src/node/proxy-server.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /**
+   * Uses the internal `NodeRadioController` to fulfill the send request.
+   * Formats the response and sends it across the socket.
+   * @param udpScanRequest  A UDP send request to fulfill and respond to.
+   * @param socket  The socket to respond on.
+   */
+  private async handleUdpSend(
```
proppy, line 73: that that handle UDP read as well?

chrome/dist/index.html
```diff
@@ -0,0 +1,72 @@
...
+<html>
+
+<head>
+  <meta charset="UTF-8" />
+  <title>Local Home SDK App</title>
+  <script type="module" src="main.js"></script>
+  <link rel="stylesheet" type="text/css" href="style.css">
+</head>
+
+<body>
```
proppy, line 11: what about building this a series of https://lit-element.polymer-project.org/ ?

chrome/src/browser/index.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Listen for the app bundle file picker.
+ */
+const filePicker = document.getElementById('app-file-picker');
+if (filePicker) {
+  filePicker.addEventListener('change', async event => {
+    const files = (event.target as HTMLInputElement).files;
+    if (files) {
+      // Run the javascript.
+      eval(await files[0].text());
```
proppy, line 67: should you run that in a worker?

chrome/src/browser/index.ts
```diff
@@ -0,0 +1,125 @@
...
+ * to an element in the page.
+ */
+const logElement = document.getElementById('log');
+if (logElement) {
+  const modifiedConsole: any = console;
+  modifiedConsole.defaultlog = console.log;
+  modifiedConsole.log = (message: string) => {
+    modifiedConsole.defaultlog(message);
+    if (logElement) {
+      logElement.innerHTML += `[${new Date().toLocaleTimeString()}]:<br/>${message}<br/>`;
```
proppy, line 20: what about a ul?

chrome/dist/index.html
```diff
@@ -0,0 +1,72 @@
...
+      <label for="udp-broadcast-port"><b>Broadcast Port</b></label>
+      <input type="text" id="udp-broadcast-port" required>
+
+      <label for="udp-listen-port"><b>Listen Port</b></label>
+      <input type="text" id="udp-listen-port" required>
+
+      <label for="udp-discovery-packet"><b>Discovery Packet</b></label>
+      <input type="text" id="udp-discovery-packet" required>
+
+      <button type="Start Scan" class="submit-button">Submit</button>
```
ToniCorinne, line 40: For a11y, consider naming the submit buttons more explicitly. "Start Scan" vs "Sent Intent" - makes it easier to navigate the page with a screen reader.

chrome/src/common/radio-message.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.udpScanConfig = udpScanConfig;
+  }
+}
+
+/**
+ * A class which contains all parameters required
+ * to fulfill a UDP send command.
+ */
+export class UdpSendRequest implements ProxyRequest {
+  proxyMessageType: RadioMessageType = 'UDPSEND';
```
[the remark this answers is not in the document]
cdaly333: The difference between `UdpRequestData` and this class is that the information in this class is independent of a `DeviceManager` instance, and represents a UDP command that could be fulfilled without needing `address` resolution.

The idea is that the `ProxyRadioServer` knows the bare minimum to fulfill the request. My instinct is that if we count on `ProxyRadioServer` to resolve the `address`, we start to share the state over both Chrome and the Node, which may make things a bit fuzzy.  The `ProxyRadioServer` is currently stateless, which I think is a good feature we may not want to give up.

On the other hand, if we take your suggestion from #48 and fold `RadioController` into `DeviceManager`, then there would be the option of moving the entire `DeviceManager` stub to the Node side.  That way we could send the entire `UdpRequestData` over, and reuse `requestId`.  

Again though, I'm not sure if this is drifting from the original goal of running the stubs in Chrome. 
 Also, having logic and state synchronize across both Chrome and Node JS runtime environments does seem a bit chaotic.

My opinion would be to move forward with the current transport classes for this PR.  I opened an issue for discussion and consideration: #61

chrome/src/node/proxy-server.ts
cdaly333: Since there is no runtime interface to configure this default, I removed it as a parameter and added a global to make the `4321` port common.  There is the odd chance port `4321` is not available, so I opened issue #62.

chrome/src/node/proxy-server.ts
cdaly333: Yes, since UDP read and write are both grouped as one command type in the platform, I just defaulted to `UdpSend` for naming.

chrome/dist/index.html
cdaly333: I opened issue #63 since I think that's a great idea but want to limit the scope of this PR.

chrome/src/browser/index.ts
cdaly333: I don't have a strong opinion whether we should or not.  This should happen almost instantly, and we don't need a reference to the app after we get the handlers.  On the other hand the `eval` call seems a bit out of place here.  Maybe @Fleker would have an opinion.

chrome/src/browser/index.ts
cdaly333: This could be addressed with #63

chrome/README.md
```diff
@@ -0,0 +1,40 @@
+# Local Home SDK Chrome Testing Environment
+
+This is a Local Home SDK implementation that runs in the Chrome browser.
```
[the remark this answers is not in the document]
cdaly333: Good point - It could be in any browser, since it's bundled. Updated naming across the project.

chrome/dist/index.html
```diff
@@ -0,0 +1,72 @@
...
+  <script type="module" src="main.js"></script>
+  <link rel="stylesheet" type="text/css" href="style.css">
+</head>
+
+<body>
+  <div id="file-picker-wrapper">
+    <label for="app-file-picker"><b>Load App Bundle: </b></label>
+    <input id="app-file-picker" type="file">
+  </div>
+  <form action="javascript:localHomePlatform.onUdpScanButton();">
```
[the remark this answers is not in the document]
cdaly333: Done.

chrome/dist/index.html
```diff
@@ -0,0 +1,72 @@
...
+    <div class="container">
+      <hr>
+      <h2>UDP Scan</h2>
```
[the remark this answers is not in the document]
cdaly333: UDP is the only discovery protocol supported for now, but I opened #57 to track implementing the rest.

chrome/src/browser/index.ts
```diff
@@ -0,0 +1,125 @@
...
+import {PlatformCoordinator} from './platform-coordinator';
+import {UDPScanConfig} from 'local-home-testing/build/src/radio/dataflow';
+
+/**
+ * Initialize the UI DOM.
+ * Redefine all `console.log` and `console.error` calls to log
+ * to an element in the page.
+ */
+const logElement = document.getElementById('log');
+if (logElement) {
```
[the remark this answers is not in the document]
cdaly333: It should never be undefined. Replaced with `!` assertion.

chrome/dist/index.html
cdaly333: Good point. Done.

chrome/src/browser/proxy-client.ts
cdaly333: See https://github.com/googleinterns/smart-home-local/pull/56#discussion_r474063678

chrome/src/browser/proxy-client.ts
cdaly333: Removed.

chrome/src/browser/index.ts
Fleker: Using a worker is probably the preferred method just for performance, and `eval` is typically not advised.

cdaly333: Hmm, the different global contexts is making it a bit tricky.

I can't pass the stubs between the worker thread and the main thread, since it appears objects have to be [`Transferable`](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) to be sent to/from a `WebWorker`.  This makes it difficult to set the handlers from a worker thread.

The best solution would probably be to run the `PlatformCoordinator` entirely in a worker thread, and have `index.ts` strictly handle UI.  Then parameters to the worker in messages, and have the worker respond to the messages with results.  This would be similar to how `CLI` works now.  However we'd be adding another layer to the app and would need to implement message handling, so I'm not sure it's worth for the performance boost.

Also, looking around for an alternative to `eval`, I'm not sure there are many ideal alternatives.  As I understand the main concern with `eval` is that it's running arbitrary javascript which isn't exactly safe.  There's the [FileSystem API](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem), which might(?) let us access the javascript file directly rather than loading the text, but inherently we're still loading arbitrary javascript.

I can open issues for these if needed (for discussion as well).  Not sure if there is a simple solution I'm missing.

PR #59, opened 2020-08-19
Change Scan results buffer format

cdaly333:
This is a minor change that switches format of the scan data stored in the `UDPScanResults` class. 

Previously the scan data was stored as a Node.js `Buffer`.  This PR changes that type to a hex `string`, which is more serializable and not dependent on the Node API.  This is a general improvement but also allows the `UDPScanResults` class to be serialized and sent over a WebSocket for #56.


proppy (approved):
Note that there are polyfill option for the `Buffer` class on Node.js: https://github.com/feross/buffer

cdaly333:
The polyfill you linked is actually bundled into `webpack`, and is used in #56.

The problem I was having was sending `UdpScanResults` over radio with a raw `Buffer` object caused it to get lost in translation.  Better to just send the hex string, so I figured I may as well change the type.

proppy:
Yep, IIRC with webpack this is done thru https://github.com/webpack/node-libs-browser/blob/master/package.json#L13 which target a fairly old version of the library (now 5.6.0).

Depending if all consumer will end up converting this to a buffer, you might consider doing the conversion before sending/receiving it over the websocket.

cdaly333:
Resolved merge conflicts.  This should be ready to merge.

